PR #25, opened 2021-09-24
some more warnings for unsupported gospel idiom

n-osborne:


src/core/ortac.ml
```diff
@@ -85,8 +85,25 @@ module Make (B : Backend.S) = struct
       List.filter_map
         (fun (sig_item : Tast.signature_item) ->
           match sig_item.sig_desc with
+          | Sig_val (decl, true) ->
+              W.register (W.Unsupported "ghost value not supported", decl.vd_loc);
+              None
           | Sig_val (decl, _ghost) when decl.vd_args <> [] -> value ~driver decl
           | Sig_val (decl, _ghost) -> constant ~driver decl
+          | Sig_type (_rec_flag, ty_decls, _ghost)
```
pascutto, line 93: Ghost types should probably have a different warning?

pascutto (commented):
You can remove the "not supported" part of the error message since the warnings will be rendered as `unsupported <msg>` in the end.

pascutto (approved):

PR #37, opened 2021-10-28
Compute mutability of type before processing specifications

n-osborne:
Traverse a type definition to compute whether it is an OCaml mutable type. 
This allows to have information even in the absence of specifications.

I also add some tests.

### DIFF
--- a/src/core/translate.ml
+++ b/src/core/translate.ml
@@ -37,12 +37,11 @@ module Mutability = struct
     || lsymbol ~driver rd.rd_cs
 
   let type_declaration ~driver (td : Tast.type_declaration) =
-    (* what about type t = int array ? what kind of type_decalration is it ?*)
+    Option.fold ~none:false ~some:(ty ~driver) td.td_ts.ts_alias
+    ||
     match td.td_kind with
     | Pty_abstract -> false
-    | Pty_variant cdl ->
-        List.exists (constructor_declaration ~driver) cdl
-        (* cdl is empty if defined through a tuple ??? *)
+    | Pty_variant cdl -> List.exists (constructor_declaration ~driver) cdl
     | Pty_record rd -> rec_declaration ~driver rd
     | Pty_open -> false
 end
--- a/src/core/translate.ml
+++ b/src/core/translate.ml
@@ -13,43 +13,6 @@ let term_printer text global_loc (t : Tterm.term) =
       (t.t_loc.loc_end.pos_cnum - t.t_loc.loc_start.pos_cnum)
   with Invalid_argument _ -> Fmt.str "%a" Tterm.print_term t
 
-let type_of_ty ~driver (ty : Ttypes.ty) =
-  match ty.ty_node with
-  | Tyvar a ->
-      Translated.type_ ~name:a.tv_name.id_str ~loc:a.tv_name.id_loc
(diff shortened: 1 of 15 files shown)


pascutto:
I'm a bit torn about the mutability status of `string`. I'm tempted to consider it as mutable, since the `String` module exposes the `unsafe_to_bytes` function, which lets you modify it. On the other hand this function is explicitly `unsafe`, so I'm not sure what we should do with this.
Do you have an opinion on this?

n-osborne:
The documentation says that `string` is not mutable anymore, except with a special option given to the compiler for compatibility purposes.
As it is still possible to have mutable strings, if we say they are not, we are making an affirmation about good practice. Per haps it is no the place of `ortac` to do so?

pascutto:
`string`s themselves are not mutable, but they can be converted to `bytes`, which are mutable. When using the `unsafe` convertion functions to do so, the underlying bytes are not copied, which means you can modify the `string` through the converted `byte`. e.g.
```ocaml
let s = "abc" in
let b = Bytes.unsafe_of_string s in
Bytes.set b 0 '0';
print_endline s
```
is valid without any compiler option, and prints `"0bc"`.

I guess the same kind of broken mutability can also happen with `Obj.magic` with basically any type. So maybe `ortac` can indeed ignore that, and assume that the program does not use unsafe functions (we'll have to write that somewhere in the documentation).

src/core/translate.ml
```diff
@@ -53,11 +53,15 @@ let var_of_arg ~driver arg : Translated.ocaml_var =
...
   let name = td.td_ts.ts_ident.id_str in
   let loc = td.td_loc in
-  let mutable_ = false (* XXX FIXME look at the ocaml type *) in
+  let mutable_ = Mutability.type_declaration ~driver td in
   let type_ = type_ ~name ~loc ~mutable_ ~ghost in
   let process ~type_ (spec : Tast.type_spec) =
     let term_printer = Fmt.str "%a" Tterm.print_term in
-    let mutable_ = type_.mutable_ || spec.ty_ephemeral in
+    let mutable_ =
+      max
```
pascutto, line 61: I think using the polymorphic `max` is a bit fragile here. Perhaps you can define an explicit one in `Mutability`?

src/core/types.ml
```diff
@@ -0,0 +1,66 @@
...
+    match t.ty_node with
+    | Tyvar _ -> Translated.Unknown
+    | Tyapp (ts, tyl) ->
+        List.fold_left
+          (fun acc t -> max acc (ty ~driver t))
+          (tysymbol ~driver ts) tyl
+
+  let lsymbol ~driver (ls : Tterm.lsymbol) =
+    Option.fold ~none:Translated.Unknown ~some:(ty ~driver) ls.ls_value
+    |> List.fold_right (fun t acc -> max acc (ty ~driver t)) ls.ls_args
```
pascutto, line 28: I think you could `fold_left` instead to get the tail-recursiveness (not that it matter that much with these small lists...).

pascutto (commented):
I think the `max` part is critical, apart from that it looks great!

src/core/types.ml
n-osborne: It was for the pipe.

PR #41, opened 2021-11-15
reorganize how to deal with runtimes and errors

### LINKED ISSUE #40: New architecture broke monolith frontend
- The `prelude` that takes `Ortac.Make` is not include anymore. Then, when a pre-condition is violated, the monolith frontend will have the default behaviour: reporting an error instead of telling monolith to ignore the test case.
- As the runtime is not open anymore, I don't think that the `prelude` would be any help to overwrite the reporting behaviour.
- Also, in the default frontend, it is empty. So maybe it is time to rethink the overall architecture ?

My proposition would be to get rid of the `prelude` and the `Functor` and overwrite the reporting behaviour in the frontend runtime (which would include the ortac runtime).

n-osborne:
Get rid of the Functor and the `prelude` expression.
Instead, the specific behaviour for a frontend w.r.t error reporting is defined in the runtime.

Fix #40


src/core/generate.ml
```diff
@@ -161,8 +161,12 @@ let axiom (a : Translated.axiom) =
   in
   [ [%stri let () = [%e body]] ]
 
-let structure driver : structure =
+let structure runtime driver : structure =
   (pmod_ident (lident (Drv.module_name driver)) |> include_infos |> pstr_include)
+  :: pstr_module
+       (module_binding
+          ~name:{ txt = Some "Runtime"; loc }
+          ~expr:(pmod_ident (lident runtime)))
```
pascutto, line 169: That's `module Runtime = <runtime>` right?
Perhaps `Runtime` is a bit too invasive as It prevents the user to call a module like this. Why did you move away from `Ortac_runtime`?

n-osborne: You're right. I'll fix this.

PR #42, opened 2021-11-28
Add `checks' verifications

pascutto:


src/core/translated.ml
```diff
@@ -7,6 +7,12 @@ type term = {
   translation : (expression, W.t) result;
 }
 
+type check = {
+  txt : string;
+  loc : Location.t;
+  translations : (expression * expression, W.t) result;
```
n-osborne, line 13: Maybe it could be a good idea to put a few words here stating the role of each of the expressions ?

n-osborne (approved):

src/core/translated.ml
pascutto: Done

PR #48, opened 2021-12-13
Invariants in known containers

n-osborne:
Check type invariants for what is inside:
- a list
- an array
- an option
- a ref

I had to add a field in Translation.type_ in order to have access to the argument of the type constructors.

This PR is incompatible with PR #51.


n-osborne:
Closing because it is stale.

PR #59, opened 2022-04-09
Update gospel

art-w:
Please feel free to ignore if this is not welcome!.. I'm dumb and did not realize that `gospel` was released on opam, so I was trying to make `ortac` compile with a `gospel` pinned to its git repo (but its interface since changed?). I've no idea what I'm doing so I randomly renamed all the missing Gospel modules until it compiled.

This gospel update breaks the ortac tests, because the latest version doesn't like invariants on non-private types (?).

```ocaml
type t = unit (*@ invariant true *)
```

```sh
$ gospel check test.mli
File "test.mli", line 1, characters 0-45:
Error: Invariant on non-private type t
```

I'm unsure how to fix these tests as a result : https://github.com/ocaml-gospel/ortac/blob/8c35f95da9ad44d09b1830abc0018c887db5fd89/test/suite/types/lib.mli#L1-L2 https://github.com/ocaml-gospel/ortac/blob/8c35f95da9ad44d09b1830abc0018c887db5fd89/test/generated/lib.mli#L1-L3

(I'm only sharing this PR in case it saves you a few minutes in the future, but again, feel free to close it if that's not the case!)


pascutto:
Thanks for the PR!
Ortac is indeed not tracking Gospel's `main`, but this is a deliberate choice as Gospel is a moving target that is hard to track, and we feel it's better to track the latest release for now.
It will definitely save us some time when we port it to `main` though, so thanks a lot!

pascutto:
Thanks again!

PR #61, opened 2022-07-04
Add some list functions

paulpatault:


shym:
This should have been addressed in #96, so I will close this PR.
Don’t hesitate to reopen it if I missed something.

PR #69, opened 2023-02-14
Move to a plugin architecture

n-osborne:
This PR proposes to move to a plugin architecture. Doing so will allow different dependencies for each modes.

In order to make the different ortac modes plugins, we also make them be subcommands. So we will write:

```
$ ortac <mode> <file>
```

rather than

```
$ ortac -f <mode> <file>
```

on the command line.

This has the benefit of allowing different modes to have different options (something that we need for the STM mode).

Making the modes be subcomands has the downside that default mode can't be the default subcommand (the one that is called if none is given) because the command line parser is then confusing the given file name with an unknown subcommand. So we can't write

```
$ ortac <file>
```

as we used to do for the default mode.

On a side note, maybe we should find a better and more meaningful name for this default mode which would no longer be default.

Along the way, we've also removed some unnecessary dependencies, which is always a win.


bin/dune
```diff
@@ -1,13 +1,27 @@
 (executable
  (name cli)
  (public_name ortac)
+ (modules sites cli)
  (package ortac)
  (libraries
+  dune-site
+  dune-site.plugins
+  dune-private-libs.dune-section
```
shym, line 9: I would say that having to resort to this would be a problem to report, as the dune documentation states about `implicit_transitive_deps false`:
> We recommend users experiment with this mode and report any problems.

shym (approved):
Very good job, paving nicely the way for the STM target.
:100: 

I’ve opened PR n-osborne/ortac#3 with a couple of typos and style suggestions, rather than typing them in the web UI...

shym:
The opam packages are slightly broken at the moment:
```
$ opam install ./*.opam plugins/monolith/*.opam
...
#=== ERROR while compiling ortac-monolith.~dev ================================#
...
### output ###
# File "dune", line 19, characters 2-17:
# 19 |   (ortac plugins)))
#        ^^^^^^^^^^^^^^^
# Error: The package ortac is not found
```

shym:
That might do the trick?
```
diff --git i/plugins/monolith/dune w/plugins/monolith/dune
index adf571d..2a0f5d1 100644
--- i/plugins/monolith/dune
+++ w/plugins/monolith/dune
@@ -17,3 +17,10 @@
  (libraries ortac-monolith.plugin)
  (site
   (ortac plugins)))
+
+(install
+ (files (ortac_monolith.cmxs as ortac-monolith/ortac_monolith.cmxs))
+ (package ortac-monolith)
+ (section
+  (site
+   (ortac plugins))))
```

n-osborne:
I'm not sure I see a real change of behaviour with this patch. It seems I have to run `opam install` twice on `ortac-monolith.opam` to get it installed either way (after having installed `ortac` on its own)...

Also, when `ortac-monolith` is installed, the plugin is not loaded (the subcommand is not available).

This fixes this second problem:
```
λ git diff
diff --git a/ortac.opam b/ortac.opam
index b230a88..78eccbd 100644
--- a/ortac.opam
+++ b/ortac.opam
@@ -23,6 +23,7 @@ build: [
     "@runtest" {with-test}
     "@doc" {with-doc}
   ]
+  ["dune" "install" "-p" name "--create-install-files" name]
 ]

 depends: [
```

PR #72, opened 2023-03-03
Refactoring in order to separate gospel to OCaml translation and assertion checking (default mode)

n-osborne:
As a follow up to the pluginisation of the monolith mode, we move the assertion checking to the default mode and expose the term translation function. This will allow the latter to be used in other modes (*eg* the qcheck-stm mode).

We use this opportunity to reshuffle functions into modules and rename these modules. So that it better matches their current usage.

co-authored by @shym


plugins/monolith/ortac-monolith.opam
```diff
@@ -0,0 +1,40 @@
+opam-version: "2.0"
+synopsis: "Monolith plugin got Ortac"
+description: "Monolith plugin got Ortac"
```
jmid, line 3: Drive-by-comment: `for Ortac` (in both lines 2 and 3)

plugins/monolith/ortac-monolith.opam
```diff
@@ -0,0 +1,40 @@
...
+maintainer: ["Nicolas Osborne <nicolas.osborne@tarides.cmd>"]
+authors: ["Nicolas Osborne <nicolas.osborne@tarides.cmd>"]
```
jmid, line 5: It is `.com`, no (in lines 4 and 5)?

shym: Indeed. It was fixed in #69 but missed in some rebasing of this PR it seems ;-)

n-osborne: Fixed :-)

plugins/monolith/ortac-monolith.opam
n-osborne: Fixed. Thanks :-)

PR #75, opened 2023-03-17
Fix monolith plugin

n-osborne:
The plugin was broken during the last gospel update.

This PR does two things:

1. Make the monolith plugin handle private types as abstract
2. Refactor dune files for the tests, mainly lib.mli is not the correct interface for the wrapper.ml generated by ortac-monolith


shym (approved):
The changes are good! :+1:
The commit message is a bit strange: unless I misunderstand something, “Refactor…” is not an extended explanation of “Handle…”; instead they are two separate items to one change which is to ”Fix compilation of the example using the monolith plugin”

n-osborne:
Thanks.
It was indeed an end-of-the-week commit message :sweat_smile:.
I've rewritten it to be more accurate.

PR #76, opened 2023-03-17
Accept any version of ocamlformat

shym:
OCamlFormat seems rather stable, so this avoids the annoyance of imposing a specific version for all developers.


shym:
Closing, as this is imposed by ocaml-ci.

PR #77, opened 2023-03-17
Configure `dune build` to skip the tests by default

shym:
Configure in details the default and runtest aliases


n-osborne (approved):
Nice and clean.
CI failure is unrelated so I move to merge.

PR #78, opened 2023-03-29
Make Warnings.kind an extensible type

n-osborne:
This PR proposes to make `Ortac_core.Warnings.kind` an extensible type. The rational is that plugins can come with their own warnings. This allows them to reuse the `Ortac_core` warning mechanism by extending the `kind` type, the `level` and the `pp_kind` functions.

### DIFF
--- a/src/core/warnings.ml
+++ b/src/core/warnings.ml
@@ -51,7 +51,9 @@ let pp_kind ppf = function
         name
   | _ -> raise Unkown_kind
 
-let pp ppf (k, loc) =
+let pp_param pp_kind level ppf (k, loc) =
   pf ppf "%a@\n%a@[%a@]@\n"
     (styled `Bold Location.print)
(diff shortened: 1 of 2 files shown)


src/core/warnings.ml
```diff
@@ -1,21 +1,25 @@
...
-type kind =
+type kind +=
   | Unsupported of string
   | Ghost_value of string
   | Ghost_type of string
   | Unsupported_model of string * string
   | Function_without_definition of string
   | Predicate_without_definition of string
 
+exception Unknown_warning
```
shym, line 14: That’s maybe a bit of a quibble, as we are already defining `Warnings.t` as a `Warning | Error`, but I would prefer `Unknown_kind`.

n-osborne: Yes, it would be more consistent, which is always good! :+1:

PR #79, opened 2023-04-17
Expose ocaml-of-gospel expression translation and move to a plugin architecture.

n-osborne:
We've been using this branch for a while and it is stable enough to be merged.

This PR contains modifications about:
- the new plugin architecture
- the refactoring allowing plugins to directly use the `ocaml_of_gospel` expression translation
- a generic way of implementing the gospel stdlib
- making ortac warnings extensible, so that plugins can add their own and still using the `core` mechanism
- some minor fixes about dune rules for tests and the monolith plugin that was broken at the last update of the gospel version

Each plugin brings a new subcommand to `ortac`. `ortac` without any plugin still comes with the default wrapper.


n-osborne:
CI is green, I move to merge.

PR #81, opened 2023-04-24
Reorganise the structure

shym:
This PR improves the structure of the project by:
- moving the tests of the monolith plugin into its dedicated directory, to unclutter the project and fix the package they were attached to,
- moving the opam files for that plugin to the root, so that it is a lot simpler to install (`opam pin https://...` sees only the packages declared at the root of the repository) and it is clearer what this repository contains,
- merging the `dune-project`, which is necessary due to the previous point,
- fixing quite a few dependencies that this reshuffling revealed.


n-osborne:
Thanks for this nice clean up!

PR #86, opened 2023-05-11
Add a subset of `Gospelstdlib.List` to the supported subset

### LINKED ISSUE #85: Extend supported gospel stdlib
For now, there is only integers and a part of array that is available for translation.
It would be nice to have list or sequence so that we can write specifications for containers.
The first use case for this extension (for now) is to write some tests for the plugins, so it need not to be complete and it would be nice not to wait that the gospel stdlib get in a stable state.

n-osborne:
Extend the supported subset of the gospel standard library with a subset of the List module (fix #85).
This should be enough in order to write some meaningful tests for the `qcheck-stm` plugin, as most of the time, `stm` models are lists.
There is no need to add too much of the List module, as the gospel standard library is expected to change in a near future.

### DIFF
--- a/src/core/context.ml
+++ b/src/core/context.ml
@@ -58,6 +58,17 @@ let supported_stdlib =
     ([ "Array" ], "length");
     ([ "Array" ], "get");
     ([ "Array" ], "for_all");
+    ([ "List" ], "length");
+    ([ "List" ], "hd");
+    ([ "List" ], "tl");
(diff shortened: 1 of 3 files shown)


shym (approved):
The code looks good to me!
The commit message says `fot he` instead of `for the`.

shym (approved):
The code looks good to me!
The commit message says `fot he` instead of `for the`.

PR #96, opened 2023-06-23
Rewrite the Ortac-runtime implementation of the Gospel Stdlib

shym:
This PR should be reviewed commit by commit, each commit message explains (in quite some details) what this PR brings.

TLDR:
- a lot more specifications should be translatable now!
- it provides an ascii encoding for the mixfix operators in order to be OCaml-syntax compatible,
- it completes the Ortac runtime with the functions from the OCaml Stdlib (leaving out functions or modules that are not so obvious, such as `Bag` and `Set`),
- it completes the Ortac-runtime interface with the whole Gospel Stdlib, with `alert` attributes on functions that are implemented, so that it pushes warning from the code generator stage to the compiling-the-generated-code stage, ie when the Ortac runtime is actually involved.

On that last aspect, it might make more sense, instead of assuming that the whole Gospel Stdlib is implemented in the Ortac runtime, to provide an explicit list of what is _not_ supported, when there is a good reason for that. In particular, `Order.is_pre_order` probably doesn’t make much sense for a testing tool such as Ortac. It would be more user-friendly to have the code generators skip such specifications instead of generating code that will never compile. Maybe that could be left for a further PR.

### DIFF
--- a/src/core/context.ml
+++ b/src/core/context.ml
@@ -33,44 +33,6 @@ let builtins =
     ([ "infix =" ], "(=)");
   ]
 
-let supported_stdlib =
-  [
-    ([], "prefix !");
-    ([], "infix +");
-    ([], "infix -");
-    ([], "infix *");
-    ([], "infix /");
-    ([], "mod");
-    ([], "pow");
-    ([], "logand");
-    ([], "prefix -");
-    ([], "infix >");
-    ([], "infix >=");
-    ([], "infix <");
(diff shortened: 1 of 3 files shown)


shym:
I just rebased this and `dune fmt`-ed it.
I also added the `unsupported_stdlib`, to avoid generating broken code.

n-osborne:
Thank you :smile:

PR #98, opened 2023-06-28
Add tests of generating hashtbl stm tests

n-osborne:
This PR proposes to add the generation of the stm tests for the hashtbl
library. This example hits more limitations than the previous one (array).
This allows to test the warning system.

This PR also add the sequence combinator to Reserr, as it is necessary to fail
at generating a case for a `cmd` in `arb_cmd` if we can't generate all the
arguments.

Co-authored with @shym


plugins/qcheck-stm/src/reserr.mli
```diff
@@ -39,7 +39,16 @@ val warn : W.t -> unit reserr
 val ( let* ) : 'a reserr -> ('a -> 'b reserr) -> 'b reserr
 val ( >>= ) : 'a reserr -> ('a -> 'b reserr) -> 'b reserr
 val ( and* ) : 'a reserr -> 'b reserr -> ('a * 'b) reserr
+
+val sequence : 'a reserr list -> 'a list reserr
+(** [sequence rs] returns [ok] of the list of ['a] iff there is no [error] in
```
shym, line 44: errors → error? (or “there are”)

PR #99, opened 2023-07-04
Add support for ghost functions and predicates

n-osborne:
- Add promote_opt to Reserr
- Fix pattern translation in core
- Add a ghost_function field to ir
- Store the implemented ghost functions in ir
- Translate and integrate to generated module the ghost functions
- Update tests

### DIFF
--- a/plugins/qcheck-stm/src/reserr.mli
+++ b/plugins/qcheck-stm/src/reserr.mli
@@ -49,6 +49,9 @@ val promote : 'a reserr list -> 'a list reserr
     no [errors] of level [Error] in [rs] and store the [errors] of level
     [Warning] in the warnings list *)
 
+val promote_opt : 'a reserr -> 'a option reserr
+(** [promote_opt r] is [promote] for a unique value *)
+
 val of_option : default:W.t -> 'a option -> 'a reserr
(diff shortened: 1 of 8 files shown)


plugins/qcheck-stm/src/reserr.mli
```diff
@@ -49,6 +49,9 @@ val promote : 'a reserr list -> 'a list reserr
     no [errors] of level [Error] in [rs] and store the [errors] of level
     [Warning] in the warnings list *)
 
+val promote_opt : 'a reserr -> 'a option reserr
+(** [promote_opt r] is [promote \[ r \] ] *)
```
shym, line 53: That documentation is bending types but maybe that’s close enough?
(Otherwise, this could be written as
```[promote_opt] is [promote] for a unique value```

shym (approved):
The history reads well, very good job!

plugins/qcheck-stm/src/reserr.mli
n-osborne: You're right, let'b be type correct!

shym:
LGTM, I approve again! :smile:

PR #104, opened 2023-07-11
update to the latest gospel

n-osborne:
Location has been fixed so display in error reporting is better


shym:
This is great! All the changed messages are indeed correct.
So now I wonder about all the messages still in “raw format”, such as the one on line 25 :thinking:

n-osborne:
We should wait until #299 is merged and then pin the merge commit

shym:
It could also be useful to merge also the commit d8fe39a from my fork to fix also the display of invariants.

PR #105, opened 2023-07-17
Implement the full Gospel Stdlib

shym:
Implement the `Bag` and `Set` modules of the Gospel Stdlib.
Define `BagSet` to factorise the implementations of `Bag` and `Set`.

`bag` and `set` are implemented as decreasing sorted lists where each element appears at most once, using the polymorphic comparison.
For `bag`s, the list contains pairs, where `snd x` is the number of occurrences of `fst x`, with `snd x > 0`.

Rationale: obviously, the time complexity of these implementations is not ideal, but:
- it is a purely functional implementation (so it can be used with QCheck-STM),
- it is compatible with the polymorphic equality: Ortac uses OCaml's `(=)` to implement Gospel's `(=)`, this ensures that `set`s and `bag`s can be tested for equality in the generated code,
- they are sorted in decreasing order so that the resulting order (according to the polymorphic comparison) on sets and bags is more natural.

Only `Order.is_pre_order` is not implemented, as it cannot be.

### DIFF
--- a/src/runtime/ortac_runtime.ml
+++ b/src/runtime/ortac_runtime.ml
@@ -162,8 +162,6 @@ module Gospelstdlib = struct
         (according to the polymorphic comparison) on sets and bags is more
         natural. *)
 
-  let niy x = failwith "%s is not implemented yet" x
-
   type 'a sequence = 'a list
   type 'a bag = ('a * Z.t) list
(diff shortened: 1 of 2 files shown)


n-osborne:
Just need an ocamlformat run

PR #106, opened 2023-07-18
Cmd pretty printer

### LINKED ISSUE #101: Generate printer for `cmd` in `qcheck-stm`

n-osborne:
This PR fixes #101 

A filtering of higher order values is added in the `ir_of_gospel` step as the pretty printer library does not support functional values. As `cmd`s correponding to higher order values are not generated in `arb_cmd` anywat, an other strategy would have been to let the pretty printer fail in these cases which it will never encounter. But now the user has a bit more information.

### DIFF
--- a/plugins/qcheck-stm/src/stm_of_ir.ml
+++ b/plugins/qcheck-stm/src/stm_of_ir.ml
@@ -9,13 +9,6 @@ let pat_default = ppat_construct (lident "Char") None
 let exp_default = evar "char"
 let res_default = Ident.create ~loc:Location.none "res"
 
-let show_attribute : attribute =
-  {
-    attr_name = noloc "deriving";
-    attr_payload = PStr [ [%stri show { with_path = false }] ];
-    attr_loc = Location.none;
-  }
-
 let may_raise_exception v =
   match (v.postcond.exceptional, v.postcond.checks) with
   | [], [] -> false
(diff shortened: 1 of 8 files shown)


shym (commented):
For the record, the review was taken offline.

plugins/qcheck-stm/src/stm_of_ir.ml
```diff
@@ -582,7 +585,61 @@ let cmd_type ir =
...
+  let get_name id = Option.fold ~none:"()" ~some:str_of_ident id |> evar in
+  let open Reserr in
+  let rec pp_of_ty ty : expression reserr =
+    match ty.ptyp_desc with
+    | Ptyp_tuple [ ty0; ty1 ] ->
+        let* pp0 = pp_of_ty ty0 and* pp1 = pp_of_ty ty1 in
+        ok (pexp_apply (evar "pp_pair") [ (Nolabel, pp0); (Nolabel, pp1) ])
+    | Ptyp_constr (lid, []) ->
+        let* s = munge_longident false ty lid in
+        ok (qualify_pp ("pp_" ^ s))
```
shym, line 602: This case is completely covered by the following one.

plugins/qcheck-stm/src/stm_of_ir.ml
```diff
@@ -582,7 +585,61 @@ let cmd_type ir =
     type_declaration ~name:(noloc "cmd") ~params:[] ~cstrs:[]
       ~kind:(Ptype_variant constructors) ~private_:Public ~manifest:None
   in
-  pstr_type Recursive [ { td with ptype_attributes = [ show_attribute ] } ]
+  pstr_type Recursive [ td ]
+
+let pp_cmd_case value =
+  let lhs = mk_cmd_pattern value in
+  let qualify_pp = qualify [ "Util"; "Pp" ] in
+  let get_name id = Option.fold ~none:"()" ~some:str_of_ident id |> evar in
```
shym, line 593: I’d suggest instead:
```ocaml
  let get_name = function
    | None -> eunit
    | Some id -> evar (str_of_ident id)
  in
```
so that we use the correct `eunit` rather than the ill-formed `evar "()"` (even if it might be pretty-printed the same way)

shym (commented):
Besides the small points I raised in the code, LGTM: very good job!

While reviewing this, I wondered whether we should add some failing tests (namely tests with bad specifications). That would allow to see the output generated by the code in this PR. Negative tests (such parallel tests on `Hashtbl`) would do.

shym (approved):
This just requires a small reformatting and it’s good to get in!

n-osborne:
Also removed the `preprocess` stanzas in the test dune file.

PR #108, opened 2023-07-19
Make wrapper a plugin

### LINKED ISSUE #100: Update documentation with wrapper plugin (was: Pluginifiy and rename default mode)
This will allow to release only the `qcheck-stm` plugin and the `ortac-core` translation.

Idea for renaming default mode is `wrapper`
We can also rename the `ortac ` parckage into `ortac-core`

Update READMEs accordingly

n-osborne:
This PR proposes to transform the default mode to the wrapper plugin.

This will allow to release different tools more independantly.

- `ortac` package is renamed as `ortac_core` as it only contains the cli and
  the `ocaml_of_gospel` translation
- `wrapper` is created and contains the code from default
- the warnings are split between `ortac_core` and `wrapper` depending on
  whether they concern the translation or the generation of runtime assertions
- an error message is displayed when the user try to use `ortac` cli whithout
  having any plugin installed

Closes #100


shym (commented):
This is nice, thank you!

I have a couple of suggestions:

1.  Running `ortac` with no argument when at least one plugin is installed, we get:
    ```
    $ deq ortac
    ortac: required COMMAND name is missing, must be 'qcheck-stm'.
    Usage: ortac COMMAND …
    Try 'ortac --help' for more information.
    [124]
    ```
    But without any plugin installed, we get:
    ```
    $ deq ortac
    ortac command-line tool requires at least one plugin.
    Please install one.[0]
    ```
    (note in particular the absence of `'\n'` at the end).
    I would prefer the two results to be as close as possible: exit code 124 and message. I suggest:

    ```diff
    diff --git i/bin/cli.ml w/bin/cli.ml
    index 8d5a18f..7de8c2e 100644
    --- i/bin/cli.ml
    +++ w/bin/cli.ml
    @@ -3,9 +3,10 @@ let () = Sites.Plugins.Plugins.load_all ()
     open Cmdliner

     let usage () =
    -  Format.(fprintf std_formatter)
    -    "ortac command-line tool requires at least one plugin.@\n\
    -     Please install one."
    +  Format.(fprintf err_formatter)
    +    "@[ortac: required plugin is missing, please install at least one@ \
    +     (qcheck-stm,@ monolith@ or@ wrapper).@]@.";
    +  exit Cmd.Exit.cli_error

     let () =
       match
    ```

2.  `Ortac_wrapper__Ir` appear many times in the `wrapper` test. I suggest giving it a short name and use that instead:
    ```diff
    diff --git i/plugins/wrapper/test/suite/translation.ml w/plugins/wrapper/test/suite/translation.ml
    index 608df6d..ac952fc 100644
    --- i/plugins/wrapper/test/suite/translation.ml
    +++ w/plugins/wrapper/test/suite/translation.ml
    @@ -1,3 +1,5 @@
    +module Ir = Ortac_wrapper__Ir
    +
     let translate path =
       let module_name = Ortac_core__Utils.module_name_of_path path in
       Gospel.Parser_frontend.parse_ocaml_gospel path
    ```
    etc.

3.  In the dune files, there is a mix up of (internal) names and public names. I would suggest always using `ortac-core` in the `libraries` fields, rather than `ortac_core`.

My fork contains a `pluginify-suggest` branch containing those suggestions.

shym:
CI seems to suggest there is something missing in the `dune` configuration, so that the plugin is not properly installed :thinking:

shym:
I’ve just added to that PR a commit that adds some explicit dependencies, to try and ensure proper ordering is happening between plugin building and tests. Let’s see what happens.

shym:
I think I (finally!) found why CI was complaining from time to time on dynlinking and found a workaround (see last commit in #109). I suppose this would be better fixed in `dune`, I plan to create a minimal example and report it upstream.

shym:
After a last review, I merge this PR, in particular to enjoy the fix for CI runs.

PR #109, opened 2023-07-26
Draft PR to test a workaround for dune in CI

shym:
This PR is just to test whether the workaround for the bug seen in CI runs during dynlink is working, as a way to test this hypothesis.


shym:
This PR served its purpose, the fix has been integrated in the merged #108.

PR #110, opened 2023-07-27
Add assymetric conjunction to possible quantification

n-osborne:
Before
```
forall x. 0 < x /\ x < 10 -> P
```
was accepted, but not
```
forall x. 0 < x && x < 10 -> P
```
This PR proposes to accept it, as I don't see any reason why it shouldn't.


shym (approved):
Thanks! Simple and to the point, and even CI likes it.

PR #111, opened 2023-07-28
Add a monolith test for the `Gospelstdlib.Set` module

shym:
Test the `Gospelstdlib.Set` module with `int` values by comparing it with the OCaml `Stdlib.Set` implementation
Use the `timeout` command to run 30 seconds of tests (but still build `stdlibset.exe` in any case)

As `Bag` shares most of its code with `Set`, this PR gives some pretty good confidence over the implementation.

### DIFF
--- a/ortac-runtime.opam
+++ b/ortac-runtime.opam
@@ -14,6 +14,7 @@ depends: [
   "dune" {>= "3.0"}
   "fmt"
   "zarith"
+  "monolith" {with-test}
 ]
 build: [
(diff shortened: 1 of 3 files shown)


shym:
I’ve updated the dune config so that it can test whether some crashes were found, _ie_ some test failed.

shym:
Rebased on `main`, now that CI tests should be fixed.

n-osborne (approved):
Nice tests :smiley: 
LGTM
CI failure can be ignored.

PR #112, opened 2023-08-22
Update READMEs (wrapper plugin, packages, Stdlib support)

### LINKED ISSUE #100: Update documentation with wrapper plugin (was: Pluginifiy and rename default mode)

shym:
Update READMEs following a series of modifications brought in various previous commits:
- the creation of the wrapper plugin,
- the reorganisation of the various packages (creation of the wrapper one, but also renaming the `ortac` into `ortac-core`),
- the full support of the Gospel Stdlib, so that the corresponding limitation can be removed from the documentation.

Closes #100


shym:
Having a quick look at the rendering, I notice that I’ve been a bit inconsistent with “Wrapper” vs “wrapper”.
Which one should we settle on?

n-osborne (approved):
Really nice! Thanks.

Just some tiny remarks:

Regarding `wrapper vs Wrapper` I would pick `wrapper` as it is not the name of
another software as `Monolith` and `QCheck_STM`.

Still on the consistency matter, `monolith/README.md` seperates `Known
limitations` and `Supported Gospel` while `qcheck-stm/README.md` does not. I
think the `qcheck-stm` version is better.

Also, in the `monolith/README.md` example the `-o` option is used while the
`wrapper/README.md` one uses a redirection. Maybe it is better to use the `-o`
option in both cases.

And finally, the core translation have this limitation that it only translates quantification on integers.

PR #116, opened 2023-09-04
Route Gospel errors as other errors in qcheck-stm plugin

shym:
Add basic support for Gospel errors in Ortac Core and use that to report properly Gospel errors in the qcheck-stm plugin


n-osborne (approved):

shym:
As CI didn’t complain (because it is offline it seems... but well...), I merged.

Issue #103, opened 2023-07-07
Rework error messages


shym:
Error messages are currently rather hard to read and interpret, and too many of them are generated.
We should:
- refine the current severity levels (warning or error), to allow for various verbosity levels,
- reword the messages,
- maybe display more context (such as the quoted extract of code).

shym:
Fixed by #122.

PR #122, opened 2023-09-07
Cleanup QCheck-STM warnings

shym:
These two parts could be seen as fairly independent but I think that using `pp_fix` has a deep impact on what we would like to see in the message: as `pp_loc` will quote the code extract, the extra information that the warning brings should usually be more about the context (such as in which function this is occurring).
So I propose to improve further this PR by modifying those, but it would be most helpful to have an example program for each of those warnings, to make sure the message does help understanding the problem.

It also contains a couple of other things to fix (`TODO` and one message that seems dead code).

### DIFF
--- a/plugins/monolith/src/warnings.ml
+++ b/plugins/monolith/src/warnings.ml
@@ -22,7 +22,7 @@ open Fmt
 let styled_list l pp = List.fold_left (fun acc x -> styled x acc) pp l
 
 let pp_level ppf = function
-  | Warning -> pf ppf "%a: " (styled_list [ `Yellow; `Bold ] string) "Warning"
+  | Warning -> pf ppf "%a: " (styled_list [ `Magenta; `Bold ] string) "Warning"
   | Error -> pf ppf "%a: " (styled_list [ `Red; `Bold ] string) "Error"
 
 let pp_kind ppf = function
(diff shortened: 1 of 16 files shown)


shym:
Note to self: this PR currently revises the QCheck-STM-specific messages, but it might be nice to ensure uniformity also with the messages coming from Ortac-core.

shym:
When ready, it will address #103.

Closes #103

n-osborne (approved):
Very nice! Thanks a lot!

Overall, error messages look way nicer and are more informative.
I've just have two tiny remarks:

We use yellow for the warnings, but OCaml uses some sort of purple. Maybe we could us the same colour.

There is a mix of `...' and `...` quotation in the error messages. I believe that the assymetric one is used when quoting the source file and the other for keywords or other expressions. I don't have strong opinion about quoting, but maybe it would be easier to respect an easier convention in the future.
A quick look shows that neither ghc nor ocaml use any quotes while gcc uses single quotation marks and clang use apostrophe.:w

shym:
I agree, I was also annoyed by the inconsistency :-)
I see no reason not to follow the style of the OCaml compiler, so I did so.
(I modified all the plugins, maybe I should have really modified only qcheck-stm?)

n-osborne:
Cool, it looks even nicer! I think it is a good idea to have modified all the plugins, consistency-wise. Eyes and CI are happy, I'll give the PR another look and it will be good to go.

PR #124, opened 2023-09-08
Seperate monadic checks and assertion.

### LINKED ISSUE #123: assertion failed about constant
Given the following interface:

```ocaml
type 'a t

val make : unit -> 'a t
(*@ t = make ()
    ensures t.contents = [] *)

val constant : unit
(*@ constant *)
```

`ortac qcheck-stm file.mli "make ()" "char t"` will complain about an internal error.

```
ortac: internal error, uncaught exception:
       File "plugins/qcheck-stm/src/ir_of_gospel.ml", line 71, characters 2-8: Assertion failed
```

n-osborne:
Tested locally, automatic tests are comming in seperated PR testing all warnings.

Fix #123

### DIFF
--- a/plugins/qcheck-stm/src/ir_of_gospel.ml
+++ b/plugins/qcheck-stm/src/ir_of_gospel.ml
@@ -184,9 +184,8 @@ let postcond spec =
 
 let val_desc config state vd =
   let open Reserr in
-  let* () = constant_test vd
-  and* () = higher_order_test vd
-  and* inst = ty_var_substitution config vd
+  let* () = constant_test vd and* () = higher_order_test vd in
+  let* inst = ty_var_substitution config vd
   and* spec =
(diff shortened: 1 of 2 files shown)


shym (approved):
Nice catch!
Just small suggestion: I didn’t understand the commit title. How about something like “Interrupt value processing earlier if conditions are not met”?

This makes me wondering whether we should use `let* ... in let* ...` rather than `and*` also in other cases. As a user, I tend to like getting as many error messages as possible on each attempt, but it might be confusing to see two messages saying that something is skipped? :thinking:

n-osborne:
I modified the commit message, which is indeed clearer.

Regarding the number of error messages (and the `let* ... in let*...` vs `let* ... and*...`), I believe we should wait to have the benefit of hindsight (I agree with both your points).

shym:
This looks good to me, and CI complains were internal failures, so I merged it.

PR #125, opened 2023-09-11
testing warnings

n-osborne:
This PR add some test for the warning system of the `ortac-qcheck-stm` plugin.
- A diff test to test the warnings of level warning (each function of the module raising a different warning).
- A cram test to test the warnings of level error

### DIFF
--- a/plugins/qcheck-stm/test/dune
+++ b/plugins/qcheck-stm/test/dune
@@ -1,3 +1,9 @@
+(cram
+ (package ortac-qcheck-stm)
+ (deps
+  (package ortac-core)
+  (package ortac-qcheck-stm)))
+
 (rule
  (target all_warnings_errors)
  (package ortac-qcheck-stm)
--- a/plugins/qcheck-stm/test/dune
+++ b/plugins/qcheck-stm/test/dune
@@ -1,3 +1,24 @@
+(rule
+ (target all_warnings_errors)
+ (package ortac-qcheck-stm)
+ (deps
+  (package ortac-core)
+  (package ortac-qcheck-stm))
+ (action
+  (setenv
+   ORTAC_ONLY_PLUGIN
+   qcheck-stm
+   (ignore-stdout
+    (with-stderr-to
+     %{target}
+     (run ortac qcheck-stm %{dep:all_warnings.mli} "make 16 'a'" "char t"))))))
+
+(rule
+ (alias runtest)
+ (package ortac-qcheck-stm)
+ (action
+  (diff all_warnings_errors.expected all_warnings_errors)))
+
 (library
  (name array)
  (modules array))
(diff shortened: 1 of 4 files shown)


plugins/qcheck-stm/test/dune
```diff
@@ -1,3 +1,6 @@
+(cram
+ (deps %{bin:ortac}))
```
shym, line 2: This should take the complete set of dependencies (`package ...`) as other tests and be attached to the qcheck-stm package.

plugins/qcheck-stm/test/dune
```diff
@@ -6,6 +9,10 @@
  (name hashtbl)
  (modules hashtbl))
 
+(library
+ (name testing_warnings)
+ (modules testing_warnings))
+
```
shym, line 15: Do we need that?

plugins/qcheck-stm/test/dune
```diff
@@ -98,6 +105,26 @@
    (diff hashtbl_errors.expected hashtbl_errors)
    (diff hashtbl_stm_tests.expected.ml hashtbl_stm_tests.ml))))
 
+(rule
```
shym, line 108: It is strange to see this inserted in the middle of the `hashtbl`-related stanzas.

plugins/qcheck-stm/test/dune
```diff
@@ -98,6 +105,26 @@
...
+ (target testing_warnings_errors)
```
shym, line 109: I suggest `all_warnings` rather than `testing_warnings` for the name of the test.

plugins/qcheck-stm/test/dune
```diff
@@ -98,6 +105,26 @@
...
+ (package ortac-qcheck-stm)
+ (deps
+  (package ortac-core)
```
shym, line 112: This should reuse the same set of dependencies and the `setenv` setting in the action than other tests.

plugins/qcheck-stm/test/testing_warnings.ml
```diff
@@ -0,0 +1,18 @@
+type 'a t = 'a array
```
shym, line 1: Do we need this, as we don’t want to compile it anyway?

shym (commented):
Nice test suite, congratulations! :tada:
Unfortunately, this reveals that there is some work left on cleaning up those error messages... :see_no_evil:
I have mostly suggestions about the `dune` configuration, to bring it on line with the other tests.

plugins/qcheck-stm/test/dune
n-osborne: github CI is indeed complaining when trying to install the packages.

plugins/qcheck-stm/test/testing_warnings.ml
n-osborne: I've removed the file, which is indeed unnecessary.

n-osborne:
Thanks for this review.
I've updated the PR accrodingly.

shym:
As it is really a quibble, I just overwrote your branch with a small difference (pushing the addition of the `cram` stanza entirely to the second commit).
Otherwise, this is good to go!

shym:
CI is happy so I merged it.

PR #126, opened 2023-09-11
Clean up opam and dune files

n-osborne:
This is an attempt to make the experimental `opam-dune-lint` CI feature happy.


shym (approved):
We took the review offline and fixed a couple of points.
Now this is good to go!

PR #127, opened 2023-09-11
Use a better work-around for the `dune-site` dependency problem for tests

shym:
The current work-around breaks `opam install --with-test ...` as the tests of each plugin depends on all plugins, from other packages (and but we really don't want to add an OPAM dependency on all plugins). So this PR introduces another work-around, inspired by a comment in ocaml/dune#8283: if `$ORTAC_ONLY_PLUGIN` is set up, load only the plugin of that name. This is using an environment variable rather than a command-line argument because the plugins provide the command-line parsers...

### DIFF
--- a/bin/cli.ml
+++ b/bin/cli.ml
@@ -1,4 +1,7 @@
-let () = Sites.Plugins.Plugins.load_all ()
+let () =
+  match Sys.getenv_opt "ORTAC_ONLY_PLUGIN" with
+  | None -> Sites.Plugins.Plugins.load_all ()
+  | Some plug -> Sites.Plugins.Plugins.load plug
 
 open Cmdliner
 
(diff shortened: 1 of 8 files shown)


n-osborne (approved):
Nice!

Just one tiny remark:
This will break if somebody decides to give a plugin a name which has not `ortac` as prefix. But this can be made a rule and we are not there yet.

shym:
> This will break if somebody decides to give a plugin a name which has not `ortac` as prefix.

Good point! I had assumed that that name was the name of the package, but it is really the name of the plugin. So I rather propose to drop the `ortac-` prefix in the plugin names. I updated the PR accordingly.

n-osborne:
CI failure is unrelated. I'm merging. Thanks!

PR #132, opened 2023-09-13
Use the core type to test the return type of `init_sut`

### LINKED ISSUE #131: qcheck-stm does not recognize non-abstract SUT types
Given the following interface `ref.mli`:

```ocaml
type t = private int ref
(*@ mutable model value : integer *)

val make : int -> t
(*@ r = make i
    ensures r.value = i *)
```

`ortac qcheck-stm ref.mli "make 42" t` complains that:

```
File "ref.mli", line 4, characters 0-126:
Error: The function `make' used for `init_state` does not return `sut`.
```

since the test is done on the Gospel type (rather than the source `core_type`) in which substitution has been done.

shym:
Use the source core type rather than the gospel type to check that the function provided by the user to initialize the SUT is indeed returning a value of the expected type

This makes the code more uniform: we are already testing the core type to detect SUT arguments. This also allows the SUT type to be concrete, as the Gospel type-checker will substitute all non-abstract types with their definitions in its AST

Closes #131

### DIFF
--- a/plugins/qcheck-stm/src/config.ml
+++ b/plugins/qcheck-stm/src/config.ml
@@ -27,15 +27,6 @@ let is_sut config ty =
   | Ptyp_constr (lid, _) -> lid.txt = sut_type_name
   | _ -> false
 
-let is_sut_gospel_ty config ty =
-  let sut_type_unqualified_name = get_sut_type_name_str config in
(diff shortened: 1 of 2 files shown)


n-osborne (approved):
Nice and clean :-) Thanks

PR #134, opened 2023-09-13
Translate Gospel `integer` as `Ortac_runtime.integer`

shym:
Translate Gospel type `integer` as `Ortac_runtime.integer` (so that the runtime can choose its implementation) rather than hard-coding it as `int`. Choosing `int` makes it inconsistent with all the implementations of the Gospel Stdlib in `Ortac_runtime` (where `Z.t` is used)


n-osborne (approved):
CI failure is unrelated.

PR #135, opened 2023-09-13
Restrict `subst_term` to `x.f` where `f` is a model field

shym:
In the QCheck-STM framework, the code generated to test whether clauses hold will be run with only the model state in scope, not the actual non-model values. `subst_term` will ensure that the substitution is indeed used in such a case instead of generating non-compiling code. This adds a new error case to report those impossible substitutions.
Add two simple tests (`ref` and record) that didn't compile before this PR.

This is based on #134 because it is using a `integer` in a model.

### DIFF
--- a/plugins/qcheck-stm/src/reserr.ml
+++ b/plugins/qcheck-stm/src/reserr.ml
@@ -24,7 +24,7 @@ type W.kind +=
   | Sut_type_not_specified of string
   | No_models of string
   | No_spec of string
-  | Impossible_term_substitution of (string * [ `New | `Old ])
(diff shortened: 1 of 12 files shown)


n-osborne (approved):
Nice catch!
And the tests make it easy to understand what's going on.

Just one remark:
The function `model_predicate` is not used anywhere.

shym:
Oups, thank you for catching the `model_predicate`, left over from a first, unnecessarily generic, draft.
Removed it and rebased.

n-osborne:
Apparently, I didn't see that it also need some formatting, sorry for the step-by-step review :sweat_smile:

shym:
The (one) CI is happy, so I merged it after reformatting.

PR #136, opened 2023-09-15
Bring Ortac up-to-date with the current development version of Gospel

shym:
This PR:
- updates the pinned gospel version, to bring its latest compatibility-breaking changes,
- bring support for patterns as arguments of anonymous functions, following ocaml-gospel/gospel#309.

As soon as gospel 0.2 is released, the pin should be dropped and converted into a regular version dependency, but this will help transition smoothly until then.

### DIFF
--- a/ortac-core.opam
+++ b/ortac-core.opam
@@ -40,5 +40,5 @@ depends: [
 ]
 
 pin-depends: [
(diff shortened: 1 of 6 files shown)


n-osborne (approved):

n-osborne:
Maybe the pinned version should include [#335](https://github.com/ocaml-gospel/gospel/pull/335) as it catches some errors at type checking before ortac can fail on them.

PR #139, opened 2023-09-19
Disable warnings in generated code

### LINKED ISSUE #138: Disable warning in generated files instead of in dune file
Adding

```ocaml
[@@@ocaml.warning -26-27-32-33-37]
```

at the beginning of the generated files would remove responsability from the user to add this in their dune files.

n-osborne:
Use a `[@@@ocaml.warning]` attribute in order to lighten the dune files
expected from the user.

Close #138


shym (approved):
Good idea, thank you!
I think some of those warnings are leftovers from intermediate development states: at the moment, our tests only trigger 26 and 27 (and with good reasons to do so) so I think it would be nicer to disable only those two. What do you think?
Otherwise, this is good to go for me.

n-osborne:
Let's remove the noise and just disable 26 and 27 then :-)

PR #141, opened 2023-09-20
Documentation/tutorial for ortac-stm plugin

n-osborne:
This PR is based on #152 in order to have to latest warning messages.

### DIFF
--- a/plugins/qcheck-stm/doc/dune
+++ b/plugins/qcheck-stm/doc/dune
@@ -1,0 +1,20 @@
+(mdx
+ (package ortac-qcheck-stm)
+ (deps
+  (package ortac-core)
+  (package ortac-qcheck-stm)))
+
+(documentation
+ (package ortac-qcheck-stm)
+ (mld_files index))
+
+(rule
+ (alias runtest)
+ (package ortac-qcheck-stm)
+ (deps
+  (:examples
+   (glob_files *.mli))
+  (package gospel))
+ (action
+  (ignore-outputs
+   (run gospel check %{examples}))))
--- a/plugins/qcheck-stm/doc/example_ghost.mli
+++ b/plugins/qcheck-stm/doc/example_ghost.mli
@@ -1,0 +1,16 @@
+type 'a t
+(*@ model size : int
+    mutable model contents : 'a list *)
+
+val make : int -> 'a -> 'a t
+(*@ t = make i a
+    checks i >= 0
+    ensures t.size = i
+    ensures t.contents = List.init i (fun j -> a) *)
+
+(* $MDX part-begin=fun-decl *)
+
+val ghost_arg : char -> 'a t -> bool
+(*@ b = ghost_arg [ i : integer] c t *)
+
+(* $MDX part-end *)
--- a/plugins/qcheck-stm/doc/index.mld
+++ b/plugins/qcheck-stm/doc/index.mld
@@ -1,0 +1,430 @@
+{0 Ortac/QCheck-STM}
+
+{1 Content}
+
+{1 Overview}
+
+The [qcheck-stm] plugin for [ortac] (called Ortac/QCheck-STM in order to avoid
+ambiguities) generates a standalone executable using
+{{:https://ocaml-multicore.github.io/multicoretests/ }QCheck-STM} to perform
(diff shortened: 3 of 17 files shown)


plugins/qcheck-stm/doc/dune
```diff
@@ -0,0 +1,19 @@
...
+
+(rule
+ (alias runtest)
+ (package ortac-qcheck-stm)
+ (deps
+  (:example example.mli)
+  (package gospel))
+ (action
+  (with-accepted-exit-codes
+   0
```
shym, line 17: Is that required? It’s the default behaviour, I think.

plugins/qcheck-stm/doc/index.mld
```diff
@@ -0,0 +1,275 @@
...
+(*@ t = make i a
+    checks i >= 0
+    ensures t.size = i
+    ensures t.contents = List.init i (fun j -> a) *)
+]}
+
+Now that [qcheck-stm] is able to generate an initial value for the type under
+test and its model, we can turn our attention to the functions we will want to
+test. Here is the example of the classic [set] function. You can see again the
+[checks] clause. Here again, it will be ignored by the tool. Then there is the
```
shym, line 68: In that case, the `checks` clause will be tested to check the function did raise an exception.

shym (commented):
This is a nice first draft, well done! I’m looking forward to seeing the complete version :smile:

jmid:
Thanks for doing this documentation. Writing nice documentation is a lot of work - so this is much appreciated! :pray: 

I suggest toning the following a bit down:
```
{2 QCheck-STM own limitations}

Finally, some limitations come directly from the QCheck-STM. There is not much
to do here except from testing the skipped function with another tool.
```

First of all, no other tools have been released so this suggestion is not helpfui.

In
> a function should have exactly one type [SUT] in its signature, and it should be as argument. 

one could support 0 or 1 arguments without hitting any QCheck-STM limitations AFAIK.

More broadly, one can encode "multiple ts" in QCheck-STM, by, say letting the SUT type be a stack of actual `t`s.
That way
- `init_sut` should initialize the stack
- `t`-consuming operations pop `t`s from the stack (a `precond` can ensure that there is enough available)
- `t`-producing operations push their result to the stack.

So this isn't a limitation of QCheck-STM as it is reasonably customizable.


> The second one is that higher order functions are not supported yet.

In multicoretests [`src/lazy/stm_tests.ml`](https://github.com/ocaml-multicore/multicoretests/blob/main/src/lazy/stm_tests.ml) contains an STM example of doing so.
It is thus a matter of emitting something along these lines from the plugin.


Let me stress again that putting in limitations to get a first version working is only natural. :100: :smiley: 
I furthermore applaud this effort to document the limitations! :tada: 

I object to listing these under `QCheck-STM own limitations` and I challenge the claim that `There is not much
to do`. 

To make this constructive, a suggestion could be to rewrite this to just say
> Please note, the current release of the plugin has the following limitations. We expect to address these in the future.

n-osborne:
Thanks for the review :-)

I was not completely happy with this part, now I have an idea on how to modify it!

plugins/qcheck-stm/doc/example_ghost.mli
```diff
@@ -0,0 +1,16 @@
...
+val make : int -> 'a -> 'a t
+(*@ t = make i a
+    checks i >= 0
+    ensures t.size = i
+    ensures t.contents = List.init i (fun j -> a) *)
+
+(* $MDX part-begin=fun-decl *)
+
+val incompatible_type : char -> string t -> bool
+(*@ b = incompatible_type c t *)
```
shym, line 14: The example is called `example_ghost.mli` but this is not what it illustrates, so I’m a bit confused here. It was overwritten with the content of example_incompatible_type, it seems.

shym (commented):
This reads very nicely, thank you!

I’ve added 4 commits containing fixes and suggestions in my `stm-plugin-documentation-plus` branch. They are separate commits to help you review my proposed changes (the one for typos should really be reviewed with `--color-words`!) but they would be better squashed in after.
I’ve also noticed that `example_ghost` doesn’t contain what it is supposed to, I think.

shym:
Thank you for the updates, LGTM. CI is happy, so I merged.

PR #144, opened 2023-09-21
Reject clauses that involves returned `sut` for `init_state`

### LINKED ISSUE #142: qcheck-stm is a bit too leniant on some clauses
Given the following source:

```ocaml
type 'a t
(*@ mutable model value : 'a list *)
val make : 'a -> 'a t
(*@ t = make a
    requires true
    ensures t.value = t.value *)
```

`ortac qcheck-stm foo.mli "make 42" "int t"` will accept to generate the following broken code:

```ocaml
let init_state = let a_1 = 42 in { value = (t_1.value) }
```
This might be making the case for a call to `subst_term` with `None` for both `old` and `new`.

n-osborne:
Close #142


shym (commented):
This is nice, thank you!
I’d rather see this PR rebased on #122, since it modifies `test_errors.t`, to make it easier to solve the conflict.
And I imagine I’ll want a new `Never` case for the `Impossible_subst` warning, since the error message we’ll see there will be far from understandable, I think.

n-osborne:
Rebased on #122.

Regarding the warning, I agree that the current error message is not very helpful (see commit message). Having slept on it, I'm wondering whether the warning should even be displayed. For example, a file like:

```
type 'a t

val make : int -> 'a -> 'a t
(*@ t = make i a
    ensures t.contents = t.contents
    ensures t.contents = List.init i (fun x -> a) *)
```

would generate a correct STM file but still display a warning about the first `ensures` clause. The function being used for INIT purpose, it is not tested anyway. So I don't see any reason why we should inform the user that we are skipping a clause. The important information is whether there is enough information to generate the `init_state` function, which will raise an error if not.

n-osborne:
Rebased on latest main.
I've also commit a new warning message.
Not completely sure about the formulation though.

shym (approved):
LGTM, thanks!

Looking for ideas for a better error message, I’m looking at what `ocamlc` reports on `type t = t` (ie “the type is cyclic”). Or using the word “recursive”? Something like “impossible to define the initial value of a model with a recursive expression”?

PR #145, opened 2023-09-21
handle returned tuple

### LINKED ISSUE #143: qcheck-stm is wrongly assuming that only one OCaml value is returned
The IR representation of values assumes that there is 0 (case of returning `unit`) or 1 returned value. But Gospel represents values returning tuples as returning multiple values (and expect the header to name each one of the members of the tuple).
See ocaml-gospel/gospel#339

So given the following source, that is valid as far as Gospel is concerned:

```ocaml
type 'a t

val make : int -> 'a -> 'a t
(*@ t = make i a
    ensures t.contents = List.init i (fun x -> a) *)

val ret_pair : 'a t -> bool * int
(*@ b, i = ret_pair t *)
```

`ortac qcheck-stm` will display:

```
ortac: internal error, uncaught exception:
       Failure("shouldn't happen (more than one return OCaml value)")
```

n-osborne:
- Handle tuple as returned value
- Remove duplicated code
- Support tuple where needed

Close #143

### DIFF
--- a/plugins/qcheck-stm/src/stm_of_ir.ml
+++ b/plugins/qcheck-stm/src/stm_of_ir.ml
@@ -460,10 +460,14 @@ let postcond_case config state idx state_ident new_state_ident value =
     in
     let pat_ret =
       match value.ret with
-      | None ->
+      | [] ->
           if may_raise_exception value then pvar (str_of_ident res_default)
           else ppat_any
-      | Some id -> pvar (str_of_ident id)
+      | [ id ] -> pvar (str_of_ident id)
+      | _ ->
+          failwith
+            "shouldn't happen (functions returning tuples are filtered out \
+             before)"
     in
     ok
       (ppat_construct (lident "Res")
@@ -487,10 +491,14 @@ let postcond_case config state idx state_ident new_state_ident value =
   in
   let res, pat_ret =
     match value.ret with
-    | None -> (evar (str_of_ident res_default), ppat_any)
-    | Some id ->
+    | [] -> (evar (str_of_ident res_default), ppat_any)
+    | [ id ] ->
         let id = str_of_ident id in
         (evar id, pvar id)
+    | _ ->
+        failwith
+          "shouldn't happen (functions returning tuples are filtered out \
+           before)"
   in
   let* rhs =
     if may_raise_exception value then
--- a/plugins/qcheck-stm/src/ir.ml
+++ b/plugins/qcheck-stm/src/ir.ml
@@ -32,7 +32,7 @@ type value = {
   sut_var : Ident.t;
   args : (Ppxlib.core_type * Ident.t option) list;
       (* arguments of unit types can be nameless *)
-  ret : Ident.t option;
+  ret : Ident.t list;
   next_state : next_state;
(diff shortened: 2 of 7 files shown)


shym (commented):
Overall, this is really nice, thank you!
The deduplication of code must be a little more subtle. I imagine we would like to:
- always use `res` as the result variable whenever `may_raise_...` is true, because, if we happen to consider a function returning a tuple, we have no one variable name to reuse,
- use the real pattern in the last `match` (either the nested one, if there might be exceptions, or the general one if not).

n-osborne:
I propose to reduce this PR to replacing internal failure by a proper warning as really handling tuples would necessitate to generate the `ty` extension with the corresponding funcitonal combinator. That would, in turn , necessitate to revise the generated module organisation (adding submodules), whiw I rather do after the release.

plugins/qcheck-stm/src/stm_of_ir.ml
```diff
@@ -460,10 +460,11 @@ let postcond_case config state idx state_ident new_state_ident value =
     in
     let pat_ret =
       match value.ret with
-      | None ->
+      | [] ->
           if may_raise_exception value then pvar (str_of_ident res_default)
           else ppat_any
-      | Some id -> pvar (str_of_ident id)
+      | [ id ] -> pvar (str_of_ident id)
+      | xs -> ppat_tuple (List.map (fun x -> pvar (str_of_ident x)) xs)
```
shym, line 467: This is currently deadcode, because `Ptyp_tuple` is not accepted in the match for `ret_ty` above, or did I miss something?
This makes it a bit harder to review. I think I would rather expect:
```ocaml
if may_raise_exception value then pvar (str_of_ident res_default)
else match value.ret with ...
```
because we don’t want to decompose right away the `result` of exception-raising functions.

If this is indeed dead code right now, it might make more sense to flag it explicitly as such, with a `failwith`, keeping all that code in a branch until we support tuples.

plugins/qcheck-stm/src/stm_of_ir.ml
```diff
@@ -487,10 +488,15 @@
   in
   let res, pat_ret =
     match value.ret with
```
shym, line 490: See remark for chunk at line 467, this would have to be modified accordingly.

shym (commented):
I agree that choosing only to report properly the fact that tuples are not supported is a better solution right now: it would be nicer to push support for tuples (up to some reasonable size, such as 10 say), would be a more elegant way to tackle this.
Maybe we could add the test of whether a tuple is returned in the `higher_order_test` function (renaming it to `no_func_arg_tuple_ret` or something similar) in `Ir_of_gospel` so that the problem is reported only once (rather than twice at the moment)?
Finally, I think it would nicer to generate an error that makes it clear that the problem is _tuples_ (`Type_not_supported "(tuple)"` could do, maybe a different error would be better to get a more specific message?).
If you agree, I can update the PR accordingly.

shym:
I just (force-)pushed an update with a specific error, that is detected during the first stage, so that we don’t produce two error messages and using `failwith` instead of dead code.

PR #149, opened 2023-09-27
Distinguish SUT-type unification errors

### LINKED ISSUE #133: Confusing display of two erros
The two different following examples have two different output. I think the second one can be confusing.

There is also a problem in the display of the first error which is labeled as a warning.

```
  $ cat > foo.mli << EOF
  > type 'a t
  > (*@ mutable model value : 'a *)
  > EOF
  $ ortac qcheck-stm foo.mli "()" "t"
  Warning: Type of system under test in `t' is incompatible with command line argument.
```

```
  $ cat > foo.mli << EOF
  > type 'a t
  > EOF
  $ ortac qcheck-stm foo.mli "()" "t"
  Warning: Type of system under test in `t' is incompatible with command line argument.
  
  File "foo.mli", line 1, characters 0-9:
  Error: The type `t' given for the system under test is not specified.
```

shym:
Unification of the SUT type is used in two different contexts:
- when handling a value, to know if its SUT-type argument has a compatible type with the one currently configured;
- when handling the declaration of the SUT type itself.
In the first case, we just want to skip the value with a warning; in the second case, we need to error out

Consequently:
- Add a new kind of error
- Feed `unify` with the information to know in which case we are

I’m a bit unsure about the phrasing of the error message, though.

Closes #133

### DIFF
--- a/plugins/qcheck-stm/src/ir_of_gospel.ml
+++ b/plugins/qcheck-stm/src/ir_of_gospel.ml
@@ -35,15 +35,20 @@ let is_a_function ty =
   let open Ppxlib in
   match ty.ptyp_desc with Ptyp_arrow (_, _, _) -> true | _ -> false
 
-let unify value_name sut_ty ty =
+let unify case sut_ty ty =
   let open Ppxlib in
   let open Reserr in
-  let show_type = Fmt.to_to_string Pprintast.core_type in
+  let fail () =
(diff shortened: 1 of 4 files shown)


n-osborne (approved):
Nice. Thank you.
I don't have any suggestion for a better error message.

shym:
As CI also was happy, I merged it then.

Issue #150, opened 2023-09-29
Issues from a first-time user


jmid:
I've just tried out the Ortac qcheck-stm plugin, taking baby steps to write a `Queue` specification.
My efforts are in https://github.com/jmid/ortac/tree/add-queue
where I mimicked the hashtbl test's files and dune rules.

When developing these I used the following invocation to get error messages
```
ortac qcheck-stm test/queue.mli "create ()" "char t" -o /tmp/queue_stm_tests.ml
```
and afterwards a `dune build @launchtests` or a direct invocation of `
_build/default/plugins/qcheck-stm/test/queue_stm_tests.exe -v` to invoke the extracted tests.

There are 4 commits in there - here's feedback for these 4 steps from yours truly: a naive, well-intending, first-time user from hell :smile: 

1. In c942852 - First version with wrong `Not_found` exception - as an end-user I was missing some information as to what call sequence caused the exception:
```
random seed: 12175242
generated error fail pass / total     time test name
[✗]    0    0    0    0 / 1000     0.0s STM Lib test sequential

--- Failure --------------------------------------------------------------------

Test STM Lib test sequential failed:

ERROR: uncaught exception in generator for test STM Lib test sequential after 1000 steps:
Exception: Failure("tl")
Backtrace: 
================================================================================
failure (1 tests failed, 0 tests errored, ran 1 tests)
```

2. In 0bde84b - Right exception, but syntax error on parenthesis - I used not unreasonable syntax (I think) which was rejected:
```
File "test/queue.mli", line 16, characters 11-12:
16 |     raises (Failure _) -> q.contents = []
                ^
Error: Gospel error: Syntax error.
```

3. In bb57690 - Right exception, but this raises Fatal error: exception Failure("tl") - I was again missing some information as to what call sequence caused the exception:
```
jmi@MintPad:~/software/ortac/plugins/qcheck-stm$ ../../_build/default/plugins/qcheck-stm/test/queue_stm_tests.exe -v
random seed: 363694356
generated error fail pass / total     time test name
[ ]    0    0    0    0 / 1000     0.0s STM Lib test sequential (generating)Fatal error: exception Failure("tl")
```

4. In 4263acb - This yields a warning about contents declared as modified, but with no translatable ensures clause... - I tried rewritting my spec into something more pure, but was missing feedback as to why the suggested spec was not translatable.

jmid:
For 4. I realize now that running the Gospel tool itself might have helped me.
[...]

PR #152, opened 2023-10-09
Adapt warning message for ensures not found

n-osborne:
This PR proposes a new warning message when the `next_state` case cannot be computed with the ensures clauses present in the specifications.

This is an attempt to answer point 4 in #150

### DIFF
--- a/plugins/qcheck-stm/src/reserr.ml
+++ b/plugins/qcheck-stm/src/reserr.ml
@@ -130,9 +130,11 @@ let pp_kind ppf kind =
       pf ppf "Skipping unsupported modifies clause:@ %a" text
         "expected \"modifies x\" or \"modifies x.model\" where x is the SUT"
   | Ensures_not_found_for_next_state (f, m) ->
-      pf ppf "Skipping %s:@ model@ %s@ %a" f m text
-        "is declared as modified by the function but no translatable ensures \
-         clause was found"
+      pf ppf "Skipping %s:@ model@ %s %a@;%a%s%a" f m text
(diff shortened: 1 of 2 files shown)


jmid (approved):
Having been confused by the original error message I think this is a nice improvement! :tada:

plugins/qcheck-stm/src/reserr.ml
```diff
@@ -130,9 +130,13 @@ let pp_kind ppf kind =
       pf ppf "Skipping unsupported modifies clause:@ %a" text
         "expected \"modifies x\" or \"modifies x.model\" where x is the SUT"
   | Ensures_not_found_for_next_state (f, m) ->
-      pf ppf "Skipping %s:@ model@ %s@ %a" f m text
-        "is declared as modified by the function but no translatable ensures \
-         clause was found"
+      pf ppf "@[Skipping %s:@ model@ %s@ %a@]@ @[%a@]@ @[%s = f (old %s)@]" f m
```
shym, line 133: I would suggest something like:
```ocaml
"Skipping %s:@ model@ %s %a@;%a@ x.%s = ..."
```
to make sure it’s clear that the `t` must appear, not `contents` by itself, and that the clause doesn’t _need_ to use `old`.

shym (approved):
Nice improvement, thanks!
I suggest a slight reformulation which might be clearer.

shym:
I (force-)pushed a proposed reformulation.

n-osborne:
Thanks. I like the error message better now.

PR #156, opened 2023-10-13
Explore conjunctive clauses

### LINKED ISSUE #147: conjunctive clauses are discarded
Currently, we look at the head of a clause (the first node in the term AST) to decide whether it contains the infomration we are looking for. This will disqualify any clause constructed as a conjunction.

For example, the following file:

```ocaml
type 'a t

val make : int -> 'a -> 'a t
(*@ t = make i a
    ensures t.contents = List.init i (fun j -> a) *)

val set : 'a t -> int -> 'a -> unit
(*@ set t i a
    modifies t.contents
    ensures true /\ t.contents = List.mapi (fun j x -> if j = (i : integer) then a else x) (old t.contents) *)
```

will raise:

```sh
$ ortac qcheck-stm file.mli "make 42 0" "int t" -o /dev/null
File "file.mli", line 10, characters 13-23:
10 |     modifies t.contents
                  ^^^^^^^^^^
Warning: Skipping set: model contents is declared as modified by the function
         but no translatable ensures clause was found.

```

This is a limitation for the generation of `next_state` and `init_state` that should be easy to overcome.

n-osborne:
Closes #147

### DIFF
--- a/plugins/qcheck-stm/src/ir_of_gospel.ml
+++ b/plugins/qcheck-stm/src/ir_of_gospel.ml
@@ -132,13 +132,15 @@ let split_args config vd args =
 
 let get_state_description_with_index is_t state spec =
   let open Tterm in
-  let pred i t =
+  let rec pred i t =
     match t.t_node with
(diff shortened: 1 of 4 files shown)


shym (approved):
LGTM, thanks!

Issue #153, opened 2023-10-10
catch exception when evaluating translated Gospel terms


n-osborne:
Some function in the Gospel standard library are not total. This means that there implementation in the `Ortac_runtime` module can raises exception, causing the evaluation of translated Gospel terms to fail.

`Ortac_runtime` also provides a `Specification_failure` exception that `wrapper` use to inform the user that the evaluation of a Gospel term has failed. This could be extended and used in `qcheck-stm`.

Another idea would be, as QCheck-STM expect `postcond` and `precond` to be pure, to catch the exceptions raised when evaluating a translated Gospel term (at least in those functions) and interpret that as the term being evaluated to `false`. `List.hd [] = a` would then be false.

shym:
Interesting ideas!
I think returning `false` would send a misleading signal, though: in the case of partial functions being used in the specification, we know that the specification is at fault, while `false` signals an inconsistency between the code and the specification. So we might guide the user in the right direction right away.
I like the idea of reusing the strategy of `wrapper` and wrapping translations of Gospel specifications in `try ... catch` blocks because we could exploit the fact that we have a little bit more context (in particular the _source_ location!).
I would favour using a dedicated exception for that purpose rather than trying to reuse the `Specification_failure` error there, as it doesn’t quite fit, I think.

n-osborne:
Closed by #157

PR #157, opened 2023-10-19
Add a Partial_specification exception to the runtime and use it in the QCheck-STM plugin

shym:
This PR improves the error message when partial functions of the runtime, ie the ones that can raise exceptions at execution time, are called out of their definition domain in the QCheck-STM plugin.

For instance, running into a `List.hd []` during the test run will now display something like:

```
exception Partial function in specification
File "array.mli", line 40, characters 12-26:
A partial function used in the specification was called out of its definition
domain, raising the following exception:
Failure("hd")
```

which points to the clause where the `Failure` was raised.

This PR thus adds:
- a `Partial_specification` exception to the runtime,
- a `term_with_catch` function that wraps the translation of a term in order to raise the `Partial_specification` exception if an exception is raised in the execution of a specification.

Hopefully this improves #150.


shym:
Closes #153.

n-osborne (approved):
Great, thanks!

I think I would rather call the new exception `Partial_function` than `Partial_specification` as the second can be a bit misleading.

CI complains about formatting.

There are some Ortac/QCheck-STM specifics in the addition to `ortac-core`, but I guess that can be generalized in the future if it need to be.

shym:
> I think I would rather call the new exception `Partial_function` than `Partial_specification` as the second can be a bit misleading.

Indeed, in the message I wrote “Partial function in a spec...”, I don’t know why I went for something different in the name... :sweat_smile:

> There are some Ortac/QCheck-STM specifics in the addition to `ortac-core`, but I guess that can be generalized in the future if it need to be.

My feeling was that the function `term_with_catch` is currently used only by Ortac/QCheck-STM, but not really specific to it, which is why I placed it there. Wrapper contains a function fairly similar, it might be possible to rewrite them so that they share it.

PR #158, opened 2023-10-19
Translate types of the Gospelstdlib

### LINKED ISSUE #155: Need to qualify type constructor in state type declaration
When declaring the type of the state, we use the type of the models.
They are usually types from the Gospelstdlib (but this is not a gospel obligation).

The code generated does not qualify the type constructors, so `list` is really from the OCaml standard library and `sequence` raises an error.

There are two possibilities:
- qualify the type constructors in the `type state` declaration (this need checking that this is indeed a type from the Gospelstdlib)
- open `Ortac_runtime.Gospelstdlib` before (we need to chack that this does not raise name capture problem)

shym:
Add a `tystdlib` to the Context, similar to the `stdlib` but for types
Use `tystdlib` in `Ocaml_of_gospel` functions translating types
Adapt QCheck-STM to feed the context to those translation functions
Add a new test for QCheck-STM using the `sequence` type for the model

Closes #155

### DIFF
--- a/plugins/qcheck-stm/src/ir_of_gospel.ml
+++ b/plugins/qcheck-stm/src/ir_of_gospel.ml
@@ -245,7 +245,7 @@ let state config sigs =
   let open Ortac_core in
   let* subst =
     Fun.flip List.assoc_opt
-    <$> (Ocaml_of_gospel.core_type_of_tysymbol ty.td_ts
+    <$> (Ocaml_of_gospel.core_type_of_tysymbol ~context:config.context ty.td_ts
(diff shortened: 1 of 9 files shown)


n-osborne (approved):
Nice, thanks!

`n-osborne/ortac#tystdlib-param` proposes a small refactoring.
It reduces the duplicated code but may not be as readable as I think.
Feel free to ignore.

shym:
Thank you! I agree with the proposed improvement, and even dropped the `fold_X_namespace` specialized functions altogether.

PR #161, opened 2023-10-23
Add a worflow to deploy documentation

n-osborne:


shym (commented):
Very good idea, thank you!
I pushed on `shym/publish-doc-mods` two small fixes:
- the cache is already performed by `setup-ocaml`, no need to enable it by hand,
- `setup-ocaml/deploy-doc` suggests to set the `permissions`, so I added them.

shym (commented):
Very good idea, thank you!
I pushed on `shym/publish-doc-mods` two small fixes:
- the cache is already performed by `setup-ocaml`, no need to enable it by hand,
- `setup-ocaml/deploy-doc` suggests to set the `permissions`, so I added them.

PR #162, opened 2023-10-23
Display the name of the tested module in the test run

shym:
Also drop the explicit `sequential`, so that users only have to change `STM_sequential` into `STM_domain` if they want to test their module in a parallel setting

This makes the display of `dune build @launchtests` a bit more useful.


n-osborne (approved):
Good idea, thanks!

PR #163, opened 2023-10-23
Generate the boiler-plate dune config for QCheck-STM tests

shym:
Generate the boiler-plate dune config for QCheck-STM tests
Also complete the `conjunctive_clauses` test so that it can be run

This PR is split up into two commits, the first one is only splitting a `dune` file into two separate files, so that it should be easy to review ignoring the moves.


n-osborne (approved):
LGTM

Thanks for splitting the PR, making the review less noisy :smiley: 

Just lacking a `dune promote` I believe.

shym:
Rebased, after the introduction of `sequence_model`.

PR #170, opened 2023-11-02
Update the opam descriptions and dunify them

### LINKED ISSUE #169: Delegate `opam` files generation to `dune`
After finding the problem that the `opam` files don’t install properly because they didn’t set up the path to the plugins directory, it might make sense to let `dune` configure this as it sees fit and move the whole generation of the opam files into dune.

shym:
The opam packages as generated by `dune` are reasonably close to hand-written ones, so use `dune` to generate them so that we are sure the `build` instructions are correct.

Closes #169


n-osborne (approved):
Thanks. LGTM.

shym:
I suppose the `lint` error in CI is due to `opam-dune-lint` not supporting the dune `"\>` syntax.
The other errors seem to be cancellations.
So I merge.

PR #174, opened 2023-11-06
Add a guard testing equality for integer constant pattern

### LINKED ISSUE #172: Problem with coercion int to integer in pattern matching
Here, gospel check succeed because `0` is coerced to an integer.

```ocaml
type 'a t
(*@ mutable model contents : 'a sequence *)

val create : unit -> 'a t
(*@ q = create ()
    ensures q.contents = Sequence.empty *)

val pop_opt : 'a t -> 'a option
(*@ o = pop_opt q
    modifies q.contents
    ensures q.contents = match Sequence.length (old q.contents) with
                          | 0 -> Sequence.empty
                          | _ -> Sequence.tl (old q.contents) *)
```

But, in the generated code, `0` is an int:

```ocaml
    let next_state cmd__002_ state__003_ =
      match cmd__002_ with
      | Pop_opt ->
          {
            contents =
              ((try
                  match Ortac_runtime.Gospelstdlib.Sequence.length
                          state__003_.contents
                  with
                  | 0 -> Ortac_runtime.Gospelstdlib.Sequence.empty
                  | _ ->
                      Ortac_runtime.Gospelstdlib.Sequence.tl
                        state__003_.contents
                with
                | e -> ...
          }
```

n-osborne:
Fixes #172

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,1 +1,8 @@
-# unreleased
+# Unreleased
+
+- Translate constant integer patterns with a guard testing for equality
+  [\#174](https://github.com/ocaml-gospel/ortac/pull/174)
+
+# 0.1.0
+
+- Initial release
--- a/plugins/qcheck-stm/src/stm_of_ir.ml
+++ b/plugins/qcheck-stm/src/stm_of_ir.ml
@@ -14,20 +14,6 @@ let may_raise_exception v =
   | [], [] -> false
   | _, _ -> true
 
-let list_fold_right1 op v xs =
-  let rec aux = function
-    | [ x ] -> x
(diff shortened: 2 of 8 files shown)


shym (commented):
Good idea, thanks!
I think it could be improved by applying the filter in the recursive descent in the pattern, otherwise something like `Some 0` will escape detection. Unfortunately, that means it must be a lot more tedious because there can be more than one such constant and all those new guards must be reported back along the recursive calls.

shym (approved):
Very nice, thank you!
I have a very minor remark: `formulae` vs `formulas` may be controversial, `formulaes` is uncontroversially wrong :-)

PR #175, opened 2023-11-07
Check for out of scope variables.

### LINKED ISSUE #171: qcheck-stm generates `next_state` function pattern matching on out of scope returned value
In the following example, which gospel typecheck, we mention the returned value in the postcondition describing the new state.

```ocaml
type 'a t
(*@ mutable model contents : 'a sequence *)

val create : unit -> 'a t
(*@ q = create ()
    ensures q.contents = Sequence.empty *)

val pop_opt : 'a t -> 'a option
(*@ o = pop_opt q
    modifies q.contents
    ensures q.contents = match o with
                          | None -> Sequence.empty
                          | Some _ -> Sequence.tl (old q.contents) *)
```

```sh
$ ortac qcheck-stm issue.mli "create ()" "int t"
```

will happily generate a next-state function pattern matching on the returned value that is not in scope:

```
    let next_state cmd__002_ state__003_ =
      match cmd__002_ with
      | Pop_opt ->
          {
            contents =
              ((try
                  match o with
                  | None -> Ortac_runtime.Gospelstdlib.Sequence.empty
                  | Some _ ->
                      Ortac_runtime.Gospelstdlib.Sequence.tl
                        state__003_.contents
                with
                | e ->
                    raise
                      (Ortac_runtime.Partial_function
                         (e,
                           {
                             Ortac_runtime.start =
                               {
                                 pos_fname = "issue.mli";
                                 pos_lnum = 11;
                                 pos_bol = 430;
                                 pos_cnum = 455
                               };
                             Ortac_runtime.stop =
                               {
                                 pos_fname = "issue.mli";
                                 pos_lnum = 13;
                                 pos_bol = 519;
                                 pos_cnum = 585
                               }
                           }))))
          }
```

n-osborne:
Fixes #171

This PR proposes to check for out of scope variables (returned values when building `next_state` function) in the `subst_term` function.

It can be argued that `subst_term` is now doing a lot of different things, but the added feature does not add lot of complexity, I used an optional arguement so that the change is transparent for user that does not need the new feature, and writing another function implementing this check would necessitate the same traversal of the term type.

Non compiling code is not generated anymore, but if the `next_state` function has not been generated, the user stay unaware of the why.

I didn't fix this last problem in this PR because it may necessitate a larger reflexion on which errors we want to display and which errors we want to forget about.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Check for out of scope variables
+  [\#175](https://github.com/ocaml-gospel/ortac/pull/175)
 - Translate constant integer patterns with a guard testing for equality
   [\#174](https://github.com/ocaml-gospel/ortac/pull/174)
 
--- a/plugins/qcheck-stm/doc/index.mld
+++ b/plugins/qcheck-stm/doc/index.mld
@@ -228,7 +228,8 @@ File "example_next_state.mli", line 15, characters 13-23:
 Warning: Skipping ensures_not_found_for_next_state: model contents is
          declared as modified by the function but no suitable ensures clause
          was found. Specifications should contain at least one "ensures
-         x.contents = ..." where x is the SUT.
+         x.contents = expr" where x is the SUT and expr can refer to the SUT
+         only under an old operator and can't refer to the returned value.
 ]}
 
 Note that you don't have to rewrite the clause. Maybe it contains information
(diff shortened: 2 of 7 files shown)


shym (approved):
Very good, thank you!
I only saw a “occurences” (instead of “occurrences”) :-)

I agree that we need to think a bit more about how to inform users about why the clause was not used to generate `next_state` without drowning them under useless warnings.

PR #176, opened 2023-11-08
Add a CI workflow for changelog

n-osborne:
Just a reminder to complete the CHANGES.md file.

Shamelessly stolen from gospel repository.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,21 @@
+name: changelog
+
+on:
+  pull_request:
+    branches: [ main ]
+    types: [ opened, synchronize, reopened, labeled, unlabeled ]
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


n-osborne:
comment.md

n-osborne:
Maybe, as there are multiple packages in the repo, we should indicate which package is concerned in the CHANGES.md entries. Or is it overkill?

shym (approved):
Good idea! LGTM

> Maybe, as there are multiple packages in the repo, we should indicate which package is concerned in the CHANGES.md entries.

I’d say that would be a good idea, in the same way we have started (maybe not that systematically...) to tag QCheck-STM-specific issues as such. Maybe with package-specific sub-sections in the changelog? But I don’t think we’d like the workflow to check anything so constraining, so let’s merge it as it is.

PR #177, opened 2023-11-09
Initiate ortac-example package

n-osborne:
The ortac-examples package aims at centralising specifications and ortac-generated code based on those specifications.

The first added example is the `lwt_dllist` librarie. This example display how to handle extension of the `STM.ty` type.

While writing the specification, I've stumbled upon #171 and #172.


shym (commented):
Nice work!
I’ve put a couple of suggestions as commits on my branch `lwt-dllist-bis` (and on (n-osborne/ortac#21)).

Other suggestions:
- I would have preferred it if `ocamlformat` had been disabled on `lwt_dllist_spec.mli` to make it easier to compare with the upstream version (it would also help maintain it)
- I expected that one of the `take_*r` functions would have a specification reading `match Sequence.length (old s.contents) with 0 -> ...` to illustrate that matching on integers is now supported :-) (it would also avoid duplicate calls to `Sequence.length` in `take_opt_r`)
- the dune rule to generate the `lwt_dllist_tests` would be nice there, even if only commented (to be honest, I would suggest to post-process them with a small script to inject the hand-written part :-)

shym:
Oh! last suggestion: we are in the ortac repository, so I would prefer the directory to be named `examples`, I feel annoyed by the redundancy :-)

shym (approved):
I’ve reviewed the new changes: well done! LGTM!

PR #178, opened 2023-11-13
Require a version of mdx supporting `.mld` files

shym:
This will hopefully fix the failures from the lower-bounds checks.


n-osborne (approved):
Apparently this is not enough to make the lower bound CI happy, but still a good idea.
Thanks!

Issue #151, opened 2023-10-03
Add warning flags the command-line


n-osborne:
The warning system for the `qcheck-stm` plugin can be a bit verbose, informing the user about each skipped clause even if there is another one that contains the information it were looking for.

I propose to have three verbosity modes:

1. error only (when `ortac` have not been able to generate anything)
2. warnings that impacts the generated artefact or test coverage (when a function has been skipped because `next-state` is impossible to generate, or because there is no `sut` type in the parameters).
3. extra verbosity, displaying every warnings possible even about skipped clauses.

`2` could be the default, `1` a `--quiet` option and `3` a `--verbose` option.

PR #179, opened 2023-11-14
Add a quiet flag

n-osborne:
This PR proposes to add a quiet flag to ortac. The flag is exposed as part of `Ortac_core.Registration` so it can be used by different plugins. Then the flag is used by `ortac-qcheck-stm`.

Using a flag implies that there are only too level of verbosity for warnings (errors are always displayed when they occur). We may to add more control over the verbosity in the future, but I think we should wait for more stability in the warning system. Still, I believe the user can benefit from a quiet mode, especially when generating tests with dune.

Related to #151

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Add a quiet flag
+  [\#179](https://github.com/ocaml-gospel/ortac/pull/179)
 - Check for out of scope variables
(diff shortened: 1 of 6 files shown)


shym (approved):
This is a good idea! Indeed, I agree that a fine-grain control would be nice but we should wait for more use cases to design it on real cases rather than what we imagine might be useful.

PR #181, opened 2023-11-16
Add an include option to `qcheck-stm` cli

n-osborne:
This PR proposes to add an include option to the `qcheck-stm` cli. This will include the module passed as value of the option in the generated code. The idea is to allow the user to easily add `STM.ty` extensions and `STM.Utils.pp*`s without having to edit the generated file by hand.

This PR does not move the configuration in the included module (the `init_sut` expression and the `sut` type expression given by the user on the cli).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Add an include option to qcheck-stm cli
+  [\#181](https://github.com/ocaml-gospel/ortac/pull/181)
 - Add a quiet flag
   [\#179](https://github.com/ocaml-gospel/ortac/pull/179)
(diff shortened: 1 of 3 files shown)


shym (approved):
Good work! I’ve opened a PR with a couple of small suggestions.
I’m not sure where the `~docv` in `conv` is used if you provide another one in `info` (I didn’t see it appear in the generated help) so, if that’s indeed overridden by the `info` one, we can go the standard combinators (as I could not find how to specify the `docv` with the combinators, I wonder whether the `~docv` is really used, by the way).
Now I want to see it put into use for the `dllist` example! (I suppose that’s how you tested it anyway :-)

PR #189, opened 2023-12-12
Remove dune-site dep for ortac-wrapper

n-osborne:
Other plugins doesn't declare it and documentation doesn't say it is necessary.


shym:
LGTM.
As CI is happy, I merge.

PR #194, opened 2024-01-15
update to latest gospel

n-osborne:
Gospel PR #374 modified how type invariants are stored in the typed ast.
This PR pin the new version (pin to be removed when a new version of Gospel is
released) and update Ortac/Wrapper code accordingly.

### DIFF
--- a/ortac-wrapper.opam
+++ b/ortac-wrapper.opam
@@ -46,3 +46,6 @@ build: [
   ["dune" "install" "-p" name "--create-install-files" name]
 ]
 dev-repo: "git+https://github.com/ocaml-gospel/ortac.git"
+pin-depends: [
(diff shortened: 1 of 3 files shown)


shym (commented):
It seems that the only plugin for which the modification has any impact is `wrapper` so the pin should appear in its opam package rather than in ortac-qcheck-stm.
We should leave the dependencies as are for all the other packages.
I’ve pushed an `update-to-latest-gospel-mini` branch on my fork with those suggestions.

n-osborne:
Thanks!
Indeed, I've extracted these two commits from the branch where I make Ortac/QCheckSTM looking at type invariants but it makes more sense to pin the new version just for the package where changes occurs in this PR.

n-osborne:
Squash merge in order to avoid extra non-compiling commits on main.

PR #195, opened 2024-01-15
Make ortac load a subset of plugins

n-osborne:
This PR is based on #194, please consider only the last commit.

This PR proposes to load the installed subset of plugins.
This will be useful for testing the dune plugin, and for generating dune file using the dune plugin when testing the other ones.

The adopted convention is to separate the plugins names with a comma.


shym (approved):
This is nice indeed, well done!
I would just have been curious to see it put into use in some test :-) But, as you say, I’m not sure we have a valid use case until the `dune` plugin is available.

PR #196, opened 2024-01-24
Read gospel files

n-osborne:
This PR proposes to use the gospel (yet unmerged) new feature from [#376](https://github.com/ocaml-gospel/gospel/pull/376).

Some refactoring is done a  long the way (there may be room for more, but I was mainly interested - at least at first - in experimenting and see how to use the new features in the examples.

This PR only add reading the `.gospel` files for the `qcheck-stm` plugin, as it is an experimentation. But this allows to see how it goes with the dune rules in the `examples/` directory.

Note that this PR is also based on [#380](https://github.com/ocaml-gospel/gospel/pull/380).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # 0.3.0
 
+- Read also `.gospel` files
+  [\#196](https://github.com/ocaml-gospel/ortac/pull/196)
 - Read an optional `cleanup` function from configuration module
   [\#226](https://github.com/ocaml-gospel/ortac/pull/226)
 - Fix field access translation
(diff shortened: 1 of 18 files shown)


n-osborne:
Rebased on #218, please consider only the last six commits.

n-osborne:
This PR could also include an update of the dune plugin to make it read the gospel file.

n-osborne:
Apparently, reading from the pre-checked `.gospel` file generate incorrect code (at least for tuples).
Postponing this feature to 0.4.0 or later.

n-osborne:
Stale

Issue #186, opened 2023-11-28
Add type invariant to postcond


n-osborne:

n-osborne:
Gospel type invariants are talking about models. As we only build the model of the type `sut` (and call it `state`), we can only check invariants on `state`

n-osborne:
I believe a `sut` is available both in `precond` and `postcond`. My point was that if the library has two types, both with gospel type invraiants, there is no need to collect the invariants for the non-`sut` type as we don't have a model to check them against for this type. (My first though and experiment was to check invariants for all the types in the library).

Checking for preservation of invariants, I'm currently implementing checking them in the `postcond` function.
Regarding the initial value, as it is provided by the user in the `init_sut` value we can either wrap this value in a check for invariants, or assume that it is the user responsability. I think it is not costly enough to not check it.

jmid:
> I believe a `sut` is available both in `precond` and `postcond`. 

Nope, unfortunately not: https://ocaml-multicore.github.io/multicoretests/0.3/qcheck-stm/STM/module-type-Spec/index.html
```ocaml
val precond : cmd -> state -> bool
val postcond : cmd -> state -> res -> bool
```

...but yes it seems I misunderstood you :slightly_smiling_face:

n-osborne:
Closed by #197

PR #197, opened 2024-02-07
Add support for type invariants

n-osborne:
This PR proposes to add checks for type invariants (for sut) in:
- `postcond` in order to guarantee preservation of invariants
- `init_state` in order to check that the user-provided value respects the invariants.

The second one is done in `init_state` and not `init_sut` as Gospel invariants are supposed to be talking about the model, not the actual value.


n-osborne:
This PR is still a draft as the type invariants are placed in an `assert` in the `init_state` function. This will cause a crash. We should find a better way to communicate with the user, via QCheck for example.

n-osborne:
This should close #186

n-osborne:
Checking for type invariant regaring `init_state` is now done with `QCheck.Test.fail_report`
This allow to have a nice test failure message from QCheck.

plugins/qcheck-stm/src/ir_of_gospel.ml
```diff
@@ -383,8 +377,8 @@ let ghost_functions =
 let run sigs config =
   let open Reserr in
   let open Ir in
-  let* state = state config sigs in
+  let* state, invariants = state config sigs in
```
shym, line 380: This line reads a bit funny. I suggest renaming the `state` function as `state_and_invariants`.

plugins/qcheck-stm/test/record_stm_tests.expected.ml
```diff
@@ -159,7 +159,15 @@ module Spec =
       match cmd__010_ with | Get -> Res (int, (get sut__011_))
   end
 module STMTests = (STM_sequential.Make)(Spec)
+let wrapped_init_state () =
+  let __state__012_ = Spec.init_state in
+  if true
```
shym, line 164: This makes me wonder: should the test deciding to generate a non-trivial `wrapped_init_state` rather be whether we have some invariants to test _after translation_, rather than `Option.is_some` at the beginning?

plugins/qcheck-stm/src/stm_of_ir.ml
```diff
@@ -843,11 +895,16 @@ let stm include_ config ir =
     let descr = estring (module_name ^ " STM tests") in
     [%stri
       let _ =
+        let open QCheck in
```
shym, line 898: Is `QCheck` necessary for more than just `Test.make`? If not, I would rather go with `QCheck.Test.make`.

shym (commented):
This looks very good, well done!

Still, I managed to find a couple of suggestions :smile:

- Is the fact that the function is named `wrapped_init_state` part of a bigger plan, because, as this PR stands, it is used only to check the initial state and its result is always discarded. In the present state, I would then suggest renaming it as `check_init_state` and giving it the type `unit -> unit`. `agree_prop` could then use a simple sequence `check_init_state (); ...`.
- I noticed that `invariants_errors.expected` appears before the `invariants.mli` is created. I would suggest squashing the “Check type invariants in `init_state`” and “Add some tests” commits.

I also put smaller comments on some lines, they were harder to describe without context.

plugins/qcheck-stm/test/record_stm_tests.expected.ml
n-osborne: Good catch.

n-osborne:
Thanks for your suggestions @shym 
Reworking `check_init_state` led to some more refactoring, but I believe it is nicer now.

shym (approved):
I agree it is nicer now, thanks!

Issue #188, opened 2023-12-06
ortac-qcheck-stm: print which invariant failed and more details about the failure


edwintorok:
Qheck-stm prints something like this (where I introduced a bug deliberately to make it fail):
```
generated error fail pass / total     time test name
[✗]    1    0    1    0 / 1000     0.0s Varray_spec STM tests

--- Failure --------------------------------------------------------------------

Test Varray_spec STM tests failed (15 shrink steps):

   push_front '\192'
   pop_at 0
   pop_front


+++ Messages ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

Messages for test Varray_spec STM tests:

  Results incompatible with model

   push_front '\192' : ()
   pop_at 0 : Ok (Elt '\192')
   pop_front : Ok (Elt 'a')
================================================================================

```

I can probably guess at which function failed (it is the last one), but I don't know from that message which postcondition failed, and what the values were (real run vs model) to debug it.

The monolith plugin prints something like this (on another test) that is better, because it shows the invariant that failed:
```
Runtime error in function `mem'
  - the invariant
      `b <-> mem i bv'
    was violated in the post-state.(* @02: Failure in an observation: candidate and reference disagree. *)
          #require "monolith";;
          module Sup = Monolith.Support;;
(* @01 *) let (Ok set0) = Sup.Exn.handle (create is Ok) 21;;
(* @02 *) let (Error _) = Sup.Exn.handle (mem is Ok 15) set0;;

```

Ideally the error message should include both sides of an invariant that failed so the developer can quickly see what the actual values causing the failure were.

I know that `qcheck-stm` only allows to return a bool, however QCheck itself does have a way to report more details that I just discovered recently: https://ocaml.org/p/qcheck-core/0.21.2/doc/QCheck2/Test/index.html#val-fail_reportf
Using `QCheck2.Test.fail_reportf` you can format a string with the actual failure and quickcheck will show that instead of just the very generic 'test failed'.

I thought that perhaps using the 'wrapper' plugin might help (which does generate the `Violated_invariant` reports with proper strings), but that doesn't quite work yet:
```
File "examples/varray_spec_wrapper.ml", line 181, characters 19-22:
181 |      try insert_at t_5 i_1 x_4
                         ^^^
Error: This expression has type 'a Ortac_runtime.Gospelstdlib.sequence ref
       but an expression was expected of type 'b t
```

[...]

PR #202, opened 2024-02-15
Improve failure message

n-osborne:
This PR improve the state of #188

This PR is based on #197, please consider only the last 7 commits.

Basically, this PR makes `postcond` return the reason of the failure rather than just a boolean, so that we can print the Gospel terms being violated back to the user in case of failing test.

This lead to give another implementation of the `STM_Sequential.Make` functor in a new `Ortac_runtime_qcheck_stm` package to use this new `postcond` function and build and pass to `QCheck` the new failure message.


n-osborne:
For demo purpose, I have pushed [this branch](https://github.com/n-osborne/ortac/tree/failure-message-example) with a failing test.

You can then see (with colours):

```
λ dune build @launchtests
49 | (rule
50 |  (alias launchtests)
51 |  (action
random seed: 476198396

Test Array STM tests failed (7 shrink steps):

   length


Gospel specification violation in function length

  File "array.mli", line 7, characters 12-18:
    i = 42
  
  File "array.mli", line 8, characters 12-26:
    i = t.size + 1
  
  length : 16

 ```

shym (approved):
Nice work, congratulations!
I confess that I didn’t carefully review the _generated code_ :sweat_smile: but this looks all good to me.

Three small questions/remarks:
- I assume you went with `(++)` to get it infix in the generated code? I think I would prefer the fully-qualified call, just like what is done for all the Gospel operators. And I would suggest picking a different one, to avoid the confusion with `Gospelstdlib.(++)`.
- This is really quibbling: I noticed various uses of `@\n`, which is better avoided according to `Format`’s documentation. Did you try alternatives (thinking it might make the generated output a bit more robust)?

n-osborne:
Thanks!

> * I assume you went with `(++)` to get it infix in the generated code? I think I would prefer the fully-qualified call, just like what is done for all the Gospel operators. And I would suggest picking a different one, to avoid the confusion with `Gospelstdlib.(++)`.

Yes, I went for the infix operator to avoid things like `(append (append x y) z)` but the risk of confusion with another function is a good reason to not do that. And if we don't have the infix notation because we qualify (which is reasonable), no reason to not adopt `append`.

> * This is really quibbling: I noticed various uses of `@\n`, which is better avoided according to `Format`’s documentation. Did you try alternatives (thinking it might make the generated output a bit more robust)?

This is the end-of-line that is mostly used in the project I think. Maybe we can change it with consistency in the whole repo if we find a better candidate.

n-osborne:
CI is green, merging.
Thanks for the review @shym :-)

PR #204, opened 2024-02-22
Print runnable scenario in case of test failure

n-osborne:
This PR is the continuation of, and is based on, #202.
Please consider only the six last commits.

This PR makes the failure message print a runnable scenario with information about returned and expected values.

It uses the same mechanism as introduced in #202 for communicating with `Qcheck.fail_reportf`.

For printing the commands with their `sut` argument in the right place, only the generated `show_cmd` has been modified.

Printing the user-provided value for `init_sut` is quite straightforward, we just have to pass the information to the runtime.

Printing the expected value is a bit more complicated.
First, we look at the Gospel specifications to see if there is enough information to compute the expected returned value (some functions may be specified in a way that their contract doesn't include a computation of the returned value).
If there is, we wrap this value in a `STM.res` type and send it to the runtime through `ortac_postond` along all the other information.
It there is not, we send a `dummy` value, to inform the printer that the expected returned value is unknown.
A warning will be displayed to the user in the second case, but this will not affect whether the function is tested or not.

This PR only deals with normal behaviour, meaning that the excpetional branches in `ortac_postond` don't send an expected returned value to the runtime for now.
I believe that handling exceptional behaviour can be done in another PR (this is me trying to keep PRs short).

The `ortac_runtime` now handle this extra information to print a runnable scenario to the user in case of test failure.

For now, I've made the choice to print the function calls in an `ignore` and to print the returned value and the optional expected returned value in comments.
We've discussed elsewhere about placing the function calls in an `assert`, but we don't always have an expected returned value

If we introduce a bug in `plugins/qcheck-stm/test/array.ml`, we have now:

```
λ dune build @launchtests
49 | (rule
50 |  (alias launchtests)
51 |  (action
random seed: 515405686

Test Array STM tests failed (14 shrink steps):

   set sut (-2603392982468686691) '\190'
   to_list sut


Gospel specification violation in function to_list

  File "array.mli", line 36, characters 12-26:
    l = t.contents
  
  let sut = make 16 'a' in
  ignore (set sut (-2603392982468686691) '\190'); (* returned Error (Invalid_argument("index ou t of bounds")) *)
  ignore (to_list sut); (* returned ['a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; '\190'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a']
                           expected ['a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a' ; 'a'; 'a'; 'a'; 'a'] *)
  

[1]
```

### DIFF
--- a/plugins/qcheck-stm/src/ir_of_gospel.ml
+++ b/plugins/qcheck-stm/src/ir_of_gospel.ml
@@ -167,6 +167,25 @@ let next_state sut state spec =
   in
   ok Ir.{ formulae; modifies }
 
+(* returns the list of terms [t] such that [ret = t] appears in the [ensures] of
+   the [spec] *)
+let returned_value_description spec ret =
+  let open Tterm in
+  let is_ret vs =
+    let open Symbols in
+    Ident.equal ret vs.vs_name
+  in
+  let rec pred t =
+    match t.t_node with
+    (* match [ret = term] and return [term] *)
+    | Tapp (ls, [ { t_node = Tvar vs; _ }; right ])
+      when Symbols.(ls_equal ps_equ ls) && is_ret vs ->
+        [ Ir.term_val spec right ]
+    | Tbinop ((Tand | Tand_asym), l, r) -> pred l @ pred r
+    | _ -> []
+  in
+  List.concat_map pred spec.sp_post
+
 let postcond spec =
   let normal = List.mapi (fun i x -> (i, Ir.term_val spec x)) spec.sp_post
   and exceptional =
@@ -216,10 +235,11 @@ let val_desc config state vd =
     in
     List.map p spec.sp_ret |> sequence
   in
+  let ret_values = List.map (returned_value_description spec) ret in
   let* next_state = next_state sut state spec in
   let postcond = postcond spec in
-  Ir.value vd.vd_name vd.vd_type inst sut args ret next_state spec.sp_pre
-    postcond
+  Ir.value vd.vd_name vd.vd_type inst sut args ret ret_values next_state
+    spec.sp_pre postcond
   |> ok
 
 let sig_item config init_fct state s =
--- a/plugins/qcheck-stm/src/stm_of_ir.ml
+++ b/plugins/qcheck-stm/src/stm_of_ir.ml
@@ -9,6 +9,7 @@ let pat_default = ppat_construct (lident "Char") None
 let exp_default = evar "char"
 let res_default = Ident.create ~loc:Location.none "res"
 let list_append = list_fold_expr (qualify [ "Ortac_runtime" ] "append") "None"
+let res = lident "Res"
 
 let may_raise_exception v =
   match (v.postcond.exceptional, v.postcond.checks) with
@@ -272,7 +273,6 @@ let run_case config sut_name value =
   let lhs = mk_cmd_pattern value in
   let open Reserr in
   let* rhs =
-    let res = lident "Res" in
     let* ty_show = exp_of_core_type value.inst (Ir.get_return_type value) in
     let ty_show =
       if may_raise_exception value then
@@ -420,6 +420,30 @@ let precond config ir =
   in
   pstr_value Nonrecursive [ value_binding ~pat ~expr ] |> ok
 
+let expected_returned_value translate_postcond value =
+  let open Reserr in
+  let ty_ret = Ir.get_return_type value in
+  let ret_val =
+    match (ty_ret.ptyp_desc, value.ret_values) with
+    | Ptyp_constr ({ txt = Lident "unit"; _ }, _), _ -> Some eunit
+    | Ptyp_tuple _, _ ->
+        failwith
+          "shouldn't happen (functions returning tuples are filtered out \
+           before)"
+    | _, [ xs ] ->
+        Option.bind
+          (to_option @@ map translate_postcond xs)
+          (Fun.flip List.nth_opt 0)
+    (* type of the returned value will be checked later with a proper error *)
+    | _, _ -> None
+  in
+  let ty_show = to_option @@ exp_of_core_type value.inst ty_ret in
+  match (ty_show, ret_val) with
+  | Some ty_show, Some ret_value ->
+      let args = pexp_tuple_opt [ ty_show; ret_value ] in
+      Some (pexp_construct res args)
+  | _, _ -> None
+
 let postcond_case config state invariants idx state_ident new_state_ident value
     =
   let open Reserr in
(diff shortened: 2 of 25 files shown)


jmid:
Nice work Nicolas :smiley::tada: I particularly liked the (aligned) `returned` and `expected` results! :heart: 

I suppose we could save a few characters (the shorter output the better IMO) by using
- top-level `let`-forms
- `let _ =` instead of `ignore( ... )`:

  ```ocaml
  let sut = make 16 'a'
  let _ = set sut (-2603392982468686691) '\190' (* returned Error (Invalid_argument("index ou t of bounds")) *)
  let _ = to_list sut (* returned ['a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; '\190'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a']
                         expected ['a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a' ; 'a'; 'a'; 'a'; 'a'] *)
  ```

As a runnable program, the example seems to place itself between two chairs.
It is both informative and runnable, yet requires manual changes: 
- it will stop with an exception on line 2 before reaching the model-mismatch in line 3.
- the mismatch in line 3 needs, e.g., an `assert` to express the expectation as runnable code,
  otherwise it will just succeed silently.
- as a very minor nit I prefixed it with `open Array` to directly copy-paste it into a top-level

Short-circuit booleans means the underlying QCheck property always stops at the first model-mismatch.
I think this means that generally
- we can ignore all results up until the last line (annotating them as comments is however very informative!)
- exception raising calls underway need some kind of `try ... with` or `protect`
- only the last line needs an `assert`, e.g., by outputting a two-line thing:
  ```ocaml
  let r = to_list sut
  let _ = assert(r = ['a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a'; 'a' ; 'a'; 'a'; 'a'; 'a']
  ```

These are just my immediate thoughts, YMMV. :shrug: 
We can also leave further improvements for the summer... (hint hint :smiley:)

jmid:
Sorry if this didn't come across above:
The output is already a nice improvement over the default qcheck-stm one! :+1:

n-osborne:
Thank you for your comments :-)

I've modified the PR so that it includes an `open` of the tested module at the beginning of the ouptuted scenario.

Regarding the use of an `assert` for the failing function call, it is worth noting that we don't always have an expected value to show: specifications can be written in a way that we can know when something goes wrong but we can't compute the expected value.

That simply means that we can't guarantee to the user that we output a failing scenario for a failing test.

plugins/qcheck-stm/src/stm_of_ir.ml
```diff
@@ -676,29 +728,36 @@ let pp_cmd_case value =
...
   let* rhs =
     let name = str_of_ident value.id in
-    let* pp_args =
-      concat_map
-        (fun (ty, id) ->
+    let rec aux ty args =
+      match (ty.ptyp_desc, args) with
+      | Ptyp_arrow (_, l, r), xs when Cfg.is_sut config l ->
+          let* fmt, pps = aux r xs in
+          ok ("%s" :: fmt, estring "sut" :: pps)
```
shym, line 735: If `fmt` directly contained `sut` instead of `%s`, we could keep `pp_args` unchanged.
On the other hand, maybe your proposition is an altogether better approach, to build `fmt` and `pp_args` at the same time.

plugins/qcheck-stm/src/stm_of_ir.ml
```diff
@@ -676,29 +728,36 @@
...
-      String.concat " " ("%s" :: List.map (Fun.const "%a") value.args)
-      |> estring
+          let* pp = pp_of_ty ty and* fmt, pps = aux r xs in
+          ok
+            ( "%a" :: fmt,
+              pexp_apply pp [ (Nolabel, ebool true) ] :: get_name id :: pps )
+      | _, [] -> ok ([], [])
+      | _, _ ->
+          failwith
+            "shouldn't happen (list of arguments should be consistent with \
```
shym, line 745: I wonder if we shouldn’t create an error value `ShouldNotHappen` (in a separate PR if at all), as it appears in many already monadic contexts.

plugins/qcheck-stm/src/ir_of_gospel.ml
```diff
@@ -167,6 +167,22 @@ let next_state sut state spec =
   in
   ok Ir.{ formulae; modifies }
 
+let returned_value spec ret =
```
shym, line 170:
```suggestion
(* returns the list of terms [t] such that [ret = t] appears in the [ensures] of
the [spec] *)
let returned_value spec ret =
```

plugins/qcheck-stm/src/ir_of_gospel.ml
```diff
@@ -167,6 +167,22 @@
...
+  let open Tterm in
+  let is_ret vs =
+    let open Symbols in
+    Ident.equal ret vs.vs_name
+  in
+  let rec pred t =
+    match t.t_node with
+    | Tapp (ls, [ { t_node = Tvar vs; _ }; right ])
```
shym, line 178:
```suggestion
(* match [ret = term] and return [term] *)
| Tapp (ls, [ { t_node = Tvar vs; _ }; right ])
```
since such patterns with guards are hard to decode

plugins/qcheck-stm/src/stm_of_ir.ml
```diff
@@ -420,6 +420,30 @@
...
+    | _, [ xs ] ->
+        Option.bind
+          (to_option @@ map translate_postcond xs)
+          (Fun.flip List.nth_opt 0)
+    (* type of the returned value will be checked later with a proper error *)
+    | _, _ -> None
+  in
+  let ty_show = to_option @@ exp_of_core_type value.inst ty_ret in
+  match (ty_show, ret_val) with
+  | Some ty_show, Some ret_values ->
```
shym, line 442: `ret_values`: this is not a list, shouldn't this be singular?

plugins/qcheck-stm/src/stm_of_ir.ml
```diff
@@ -431,7 +455,25 @@
...
+    | None ->
+        let* () =
+          warn
+            ( Incomplete_ret_val_computation (Fmt.str "%a" Ident.pp value.id),
+              value.id.id_loc )
+        in
+        ok @@ esome dummy
+    | Some e -> ok @@ esome e
+  in
+  let wrap_check ?(normal = false) t e =
```
shym, line 476: What does `normal` really mean here? I see it set to `true` for (normal) postconditions and invariants, but anyway, I don’t have a clear understanding of the impact of setting it to `true`.

shym (commented):
I must say this took quite some time to review, maybe some comments could help.
I wrote my comments on specific lines except for:
- commit “Print cmds with their sut argument”: the message could mention that it also adds `open X` at the beginning of the generated log
- commit “Collect term containing a computation of the returned value”: I must admit I had to read the code to understand the commit message; I would suggest adding a small comment to the function `returned_value` (maybe it would be better named `returned_value_specification`?)

plugins/qcheck-stm/src/stm_of_ir.ml
n-osborne: We can indeed spare the `%s` and `estring "sut"`

plugins/qcheck-stm/src/stm_of_ir.ml
n-osborne: I've added some comments to explain.
Thanks!

n-osborne:
Green light to merge has been given in another chat.

PR #205, opened 2024-03-05
Add an optional argument for protecting the run of the tests

n-osborne:
This PR propose to add a `-p` optional argument to the Ortac/QCheck-STM
command-line.

The new optional argument takes a string which should be the name of the function to call.
The function should be implemented in the module passed to the `-i` option.
The funciton has type `(unit -> unit) -> unit`, which means that running the tests is suspended so the basic usage is:

```ocaml
let run_tests tests = (* do something *) (tests ())
```


n-osborne:
Note to myself: I need to add the equivalent optional argument to Ortac/Dune plugin.

n-osborne:
Converting to draft as it is not up to date with the module-based configuration from #214

n-osborne:
Stale

PR #206, opened 2024-03-06
Add support for exceptional behaviour to printing runnable scenario

### LINKED ISSUE #188: ortac-qcheck-stm: print which invariant failed and more details about the failure

n-osborne:
This PR proposes to add support for exceptional behaviour when printing a runnable scenariowhen a test is failing.

This PR is based on #204, please consider only the two last commits.

This PR should complete the story to fix #188

Again, if we introduce a bug in `plugins/qcheck-stm/test/array.ml` (here `let get a _ = get a 7`), we have now:

```
λ dune build @launchtests
49 | (rule
50 |  (alias launchtests)
51 |  (action
random seed: 163835566

Test Array STM tests failed (5 shrink steps):

   protect (fun () -> get sut (-952798241484687328))


Gospel specification violation in function get

  File "array.mli", line 11, characters 11-26:
    0 <= i < t.size
  
  open Array
  let protect f = try Ok (f ()) with e -> Error e
  let sut = make 16 'a'
  let r = protect (fun () -> get sut (-952798241484687328))
  assert (match r with
            | Error (Invalid_argument _) -> true
            | _ -> false)
  (* returned Ok ('a') *)
  

[1]
```

So the call to any function that may raise an exception (according to the Gospel psecifications) are protected to return a `('a, exn) result` instead of a simple `'a` (as it is done today in QCheck-STM).

The value returned by the protected call is then compared to the expected result: here we were expecting an `Invalid_argument` exception because the clause is a `checks`, but in the case of a `raises` clause, the name of the expected exception will be printed.

Note that only the name of the exception is printed, this PR does not propose to handle the arguments of the said exception. I may want to wait to see how [gospel#379](https://github.com/ocaml-gospel/gospel/issues/379) evolves.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,8 @@
 
 - Improve test-failure message
   [\#202](https://github.com/ocaml-gospel/ortac/pull/202) and
-  [\#204](https://github.com/ocaml-gospel/ortac/pull/204)
+  [\#204](https://github.com/ocaml-gospel/ortac/pull/204) and
+  [\#206](https://github.com/ocaml-gospel/ortac/pull/206)
 - Add a comment warning that the file is generated
   [\#198](https://github.com/ocaml-gospel/ortac/pull/198)
 - Add support for type invariants
--- a/examples/lwt_dllist_tests.ml
+++ b/examples/lwt_dllist_tests.ml
@@ -25,8 +25,8 @@ module Spec =
           Format.asprintf "%s %a sut" "add_l" (Util.Pp.pp_int true) a_1
       | Add_r a_2 ->
           Format.asprintf "%s %a sut" "add_r" (Util.Pp.pp_int true) a_2
-      | Take_l -> Format.asprintf "%s sut" "take_l"
-      | Take_r -> Format.asprintf "%s sut" "take_r"
+      | Take_l -> Format.asprintf "protect (fun () -> %s sut)" "take_l"
+      | Take_r -> Format.asprintf "protect (fun () -> %s sut)" "take_r"
       | Take_opt_l -> Format.asprintf "%s sut" "take_opt_l"
       | Take_opt_r -> Format.asprintf "%s sut" "take_opt_r"
     type nonrec state = {
--- a/examples/lwt_dllist_tests.ml
+++ b/examples/lwt_dllist_tests.ml
@@ -337,7 +337,7 @@ let ortac_postcond cmd__005_ state__006_ res__007_ =
           else
             Some
               (Ortac_runtime.report "Lwt_dllist_spec" "create ()"
-                 (Some (Res (Ortac_runtime.dummy, ()))) "is_empty"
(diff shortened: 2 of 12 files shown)


shym (approved):
This reads very nicely (might be a question of momentum, after reviewing a series of such PRs :-), thank you!

I wondered why you chose `Either` (introduced in 4.12 IIUC) rather than `Result` (introduced in 4.08), both on the dependency aspect and the fact that `Result` is a sensible use case to encode exception or not.
I would suggest we raise the dependency where needed if we keep `Either`.

By the way, I see that our runtime is already requiring 4.12 (do you remember why? the git history suggests that it was fixed thanks to a OPAM-CI complaining when we released 0.1, but it is otherwise short on details :-/ I should have written a better commit message then). If the reason for that 4.12 dependency for the runtime was strong anyway, we might just adopt it all across our packages.

PR #210, opened 2024-03-15
Fix a small bug in a cram test and some typos

shym:
This tiny PR fixes a bug with the `dune-rules` cram test which might trigger dynlink failures. It also proposes to rephrase a sentence and to fix typos.


n-osborne:
Thanks!
Nice catch

n-osborne (approved):

PR #215, opened 2024-04-08
Use `dune-build-info` rather than `dune subst`

n-osborne:
This PR proposes to move from `dune subst` to `dune-build-info` for the `--version` option.

The version number is stored in the `dune-project` file. It should be bumped before a new release.

I also propose to tag the development version with `.dev`. The idea is that if one wants more precise information about a far is the installed development version is from the last release, one can ask `opam pin` or just `git`.


shym (commented):
Thank you for that, it will make releases a lot nicer :-)

I would suggest two small modifications:
- still keep the `dune subst`, as it also allows an `ortac` installed by pinning a git repository to provide a precise version identifier, even if it doesn’t show up when running directly the binary in the working copy with `dune exec` (...),
- use `+dev` rather than `.dev`, as I think that’s closer to the convention; in version numbers, I tend to interpret (hopefully I’m right to do so :-):
  - `+something` as an indicator for “with some more”,
  - `~something` as pre-release (typically `~alpha`),
  - `-something` as a rather neutral qualifier.

Those suggestions are available in my `dune-build-info-suggest` branch if you are interested.

n-osborne:
Thanks. I didn't know that `dune subst` would change anything without watermarks. That's great!

I agree with the `+dev`, this is what shows up with `opam` anyway so I believe this is the way to go.

shym (approved):
LGTM!

n-osborne:
Thanks!
Merging.

PR #217, opened 2024-04-09
Move to use a default `Cmdliner.term` rather than the `usage` function

### LINKED ISSUE #216: `ortac` binary always requires a plugin, even to just show up its help or version number
In a just-cleaned repository,

```
$ dune exec -- ortac --help
ortac: required plugin is missing, please install at least one (qcheck-stm,
monolith or wrapper).
```

and

```
$ dune exec -- ortac --version
ortac: required plugin is missing, please install at least one (qcheck-stm,
monolith or wrapper).
```

which is all rather silly. Could we support those options, or at least rephrase the message (because in that case, the user really didn’t require any plugin)?

n-osborne:
Fixes #216

The inspiration comes from [this example](https://erratique.ch/software/cmdliner/doc/examples.html#exdarcs)
Though use of `ret` is [discouraged](https://erratique.ch/software/cmdliner/doc/Cmdliner/Term/index.html#val-ret).
But I didn't find any alteratives.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Make `--help` and `--version` work even without any plugins installed
+  [\#217](https://github.com/ocaml-gospel/ortac/pull/217)
 - Improve test-failure message
   [\#202](https://github.com/ocaml-gospel/ortac/pull/202) and
   [\#204](https://github.com/ocaml-gospel/ortac/pull/204) and
--- a/bin/cli.ml
+++ b/bin/cli.ml
@@ -5,25 +5,31 @@ let () =
 
 open Cmdliner
 
-let usage () =
-  Format.(fprintf err_formatter)
-    "@[ortac: required plugin is missing, please install at least one@ \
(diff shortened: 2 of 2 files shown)


shym (commented):
I wrote an extra commit (I propose we squash them) on my `default-term-sugg` branch to improve the error reporting and the commit message.
In particular, I really prefer keeping a small usage message when no command was given, rather than being sent directly into a pager with a man page (which gives much more information (TMI!) than just what is the problem).
We get a behaviour that’s closer to what happens if you ask for a plugin that is not installed, which is nice.
I follow your lead in dropping the list of suggested plugins, maybe it’s not that useful. And, technically, it is possible to have only the `dune` plugin, but it would be pointless.

n-osborne:
Thanks, I've squashed your commit and add you as a co-author as you end up writing about half of the modifications of this PR.

shym:
I just force-pushed a new version of the commit just to change the message: the message was misleading, as we don’t add a default `Cmd`, the default is a `Term.t`.

shym (approved):
LGTM now (quite obviously... :-)!

PR #214, opened 2024-03-28
Move to a module based configuration

n-osborne:
Fixes #173

This PR proposes to move from a command-line based configuration to a module-based configuration.

It does so adopting the proposition of [this comment](https://github.com/ocaml-gospel/ortac/issues/173#issuecomment-2020541798)

It is in two folds: the first 17 commits change the way Ortac/QCheck-STM is configured and the next 4 commits update the Ortac/Dune plugin and the examples.

The changes in the Ortac/Dune module also integrate some sort of fix: the `ortac dune qcheck-stm` command now takes the actual filenames (that is with the extensions). I find it to be more natural.

PR #218, opened 2024-04-10
Revise Ortac/Dune to minimize yet again boilerplate

n-osborne:
This PR proposes to:

1. add an optional argument for the library name. Default is the base name of the module under tested.
2. make the configuration and the generated tests filename arguments optional. Default is the base name of the module under test prefixed with [`_config` `_tests`]`.ml`

In the best case scenario, this allows to just write:

```dune
(rule
 (alias runtest)
 (mode promote)
 (targets dune.inc)
 (action
  (with-stdout-to
   %{targets}
   (run ortac dune qcheck-stm lib.mli))))

(include dune.inc)
```


n-osborne:
This PR is based on #214, lease consider only the last 5 commits.

plugins/dune-rules/src/dune_rules.ml
```diff
@@ -14,38 +14,69 @@ end = struct
...
+            ~docv:"OCAML_OUTPUT")
+
+    let library =
+      Arg.(
+        value
+        & opt (some string) None
+        & info [ "l"; "library" ]
+            ~doc:
+              "Library name. Assume it is the same as the module under test if \
+               missing."
```
shym, line 49: Did you try using the optional `~absent` argument for `Arg.info`?

plugins/dune-rules/test/test.t
```diff
@@ -65,41 +49,18 @@ generated.
 When the optional output argument is set, rules will be written in the file and will reflect
```
shym, line 49: The end of this sentence talks about the now-obsolete rule to generate dune rules, doesn’t it?

plugins/dune-rules/src/dune_rules.ml
```diff
@@ -14,38 +14,69 @@
...
     let config_file =
       Arg.(
-        required
-        & pos 1 (some string) None
-        & info [] ~doc:"Configuration file for Ortac/QCheckSTM" ~docv:"CONFIG")
+        value
+        & opt (some string) None
+        & info [ "c"; "config" ]
+            ~doc:
+              "Configuration file for Ortac/QCheckSTM Default is the \
```
shym, line 26: Missing `'.'` between the 2 parts of the documentation?

shym, line 26: `Ortac/QCheck-STM`, with a hyphen?

examples/dune
```diff
@@ -8,6 +8,18 @@
   (:standard -w -69))
  (package ortac-examples))
 
+(rule
+ (alias runtest)
+ (mode promote)
+ (targets dune.lwt_dllist.inc)
+ (action
+  (with-stdout-to
+   %{targets}
```
shym, line 17:
```suggestion
 (action
  (with-stdout-to
   dune.lwt_dllist.inc
```
The target file `with-stdout-to` writes to is automatically added to the targets, and I think the shorter version is also nicer to read.
If you adopt this suggestion, there are a couple of other occurrences in the same file and a few others elsewhere in the repository (the ones in generated rules are less interesting to change, as they are not models of rules users should write by hand).

plugins/dune-rules/README.md
```diff
@@ -20,18 +20,28 @@ plugins (the qcheck-stm plugin for now). You have to give it the option you want
...
-$ ortac dune qcheck-stm lib.mli lib_conf.ml lib_tests.ml --package=pack --with-stdout-to=dune.inc
+```dune
+(rule
+ (alias runtest)
+ (mode promote)
+ (targets dune.inc)
+ (action
+  (with-stdout-to
+   %{targets}
+   (run ortac dune qcheck-stm lib.mli))))
```
shym, line 35:
```suggestion
 (mode promote)
 (action
  (with-stdout-to
   dune.inc
   (run ortac dune qcheck-stm lib.mli))))
```

plugins/dune-rules/README.md
```diff
@@ -20,18 +20,28 @@
...
 ```
 
-to generate the dune rules to generate and run the tests. You can then include
-`dune.inc` in the `dune` file and the next time you run `dune runtest`, your
-code should be tested using QCheck-STM.
+This stanza assumes that you have written the configuration for
+[Ortac/QCheck-STM] in a file named `lib_config.ml` and that the `Lib` module is
+part of the `lib` library. It will write the generated tests in `lib_tests.ml`.
+If you want more control, you can use the `--config` `--library` and `--output`
+to give custom name to these files.
```
shym, line 44:
```suggestion
If you want more control, you can use the `--config`, `--library` and `--output`
command-line options to give custom names to these files.
```

shym (approved):
This is really nice simplification, thank you! Now I wonder if those default names should not be adopted directly in Ortac/QCheck-STM.

- 8198b4a Remove generation of dune rule generation  
  I find the phrasing of the commit message unclear, isn’t a negation missing?
  Suggestion: “as we want...” → “Future commits will simplify calling out directly the `dune qcheck-stm`” from the command line.” (even if I’m not a big fan a talking about future commits in messages, here that might be the needed context)
- 9990674 Make config and test name optional  
  Change the commit title to _names_, to make it clearer the config is not optional, just its name?
  I wondered whether it would be worth having tests for the case the options are actually working when explicitly set, for instance by using them in the examples to preserve the file names used before the PR?

plugins/dune-rules/test/test.t
n-osborne: Yes it is :-)

plugins/dune-rules/README.md
```diff
@@ -20,18 +20,27 @@ plugins (the qcheck-stm plugin for now). You have to give it the option you want
...
+(include dune.inc)
 ```
 
-to generate the dune rules to generate and run the tests. You can then include
-`dune.inc` in the `dune` file and the next time you run `dune runtest`, your
-code should be tested using QCheck-STM.
+This stanza assumes that you have written the configuration for
+[Ortac/QCheck-STM] in a file named `lib_config.ml` and that the `Lib` module is
+part of the `lib` library. It will write the generated tests in `lib_tests.ml`.
+If you want more control, you can use the `--config` `--library` and `--output`
```
shym, line 42:
```suggestion
If you want more control, you can use the `--config`, `--library` and `--output`
```

plugins/dune-rules/src/dune_rules.ml
```diff
@@ -14,38 +14,71 @@ end = struct
...
-        & info [] ~doc:"Configuration file for Ortac/QCheckSTM" ~docv:"CONFIG")
+        value
+        & opt (some string) None
+        & info [ "c"; "config" ]
+            ~doc:
+              "Configuration file for Ortac/QCheck-STM. Useful for generating \
+               multiple test files per module under test."
+            ~absent:
+              "Concatenation of INTERFACE without the extension and \
+               \"_config.ml\"."
```
shym, line 30: Given how the `~absent` field is rendered, I think I would prefer no initial capital and no ending period.

```suggestion
              "concatenation of INTERFACE without the extension and \
               \"_config.ml\""
```

shym (approved):
A small comment now that I see how `~absent` is rendered.

PR #223, opened 2024-04-16
Fix lower bounds

n-osborne:


jmid:
A suggestion:
For qcheck we've been bitten in the past by trying to support cross-versions as it quickly becomes messy.
Scenario: "Will ortac-runtime.0.5.0 require ortac-core.0.3.0 or only ortac-core.0.2.0"?
This will largely be wasted effort trying to debug lower bound failures! :grimacing: 
A qcheck example of this happening in practice: https://github.com/ocaml/opam-repository/pull/24703

For this reason, I suggest using
```
  "ortac-core" {= version}
```
etc. so that ortac-runtime.0.2.0 requires precisely ortac-core.0.2.0 (and so on)
in the interest of saving you from additional pain later :slightly_smiling_face:

ortac-qcheck-stm.opam
```diff
@@ -32,10 +32,8 @@ depends: [
   "qcheck-core" {with-test}
   "qcheck-stm" {with-test}
   "qcheck-multicoretests-util" {with-test}
-  "ortac-core"
+  "ortac-core" {= "0.2.0"}
```
jmid, line 35: Both dune and opam support the `version` variable.
A dune example here: https://github.com/ocaml-multicore/multicoretests/blob/main/dune-project#L35
A (generated) opam example here: https://github.com/ocaml-multicore/multicoretests/blob/main/qcheck-stm.opam#L27
Using it has the advantage of needing to go around and updating many version numbers on each release :slightly_smiling_face:

n-osborne: Nice!

Issue #209, opened 2024-03-12
Unbound constructor error in Ortac-generated file


Lucccyo:
Ortac generated a file that did not include the constructor of the type `A` I created in the gospel (it should do the same for type `B`).
```
(*@ type v = A of string | B of string *)
```
While running QCheckSTM tests on Ortac file, it fails with the following:
```
5 |     | Some v_1 -> (match v_1 with | A _ -> true | B _ -> false)
                                        ^^^
Error: Unbound constructor A
```

n-osborne:
The scope of this issue is in fact a bit larger than I first thought.

Indeed, all the plugins could benefit from the support for Gospel ghost types translation.

This means that some modifications should be done in `ortac-core`:

There are mainly two modifications to bring to `ortac-core`

- Provide a translation from `Gospel.Tast.type_declaration` to `Ppxlib.type_declaration`
- Store constructors and destructors of the Gospel types in the `Context.t` (in a similar way that ghost functions are stored)

I'm not sure whether we have to handle name capture as Gospel type-checker may complain at an earlier stage.

Regarding the second item, we have to deal with
- variants that take a tuple as argument
- variants that take an inlined record as argument (though they seem buggy in Gospel)
- records' fields (destructors)
- records' functional constructor that is generated by Gospel and injected in the typed AST

Once these modifications are done, we should bring the change to `ortac-qcheck-stm` and to `ortac-wrapper` (`otrac-monolith` is based on `ortac-wrapper` so this will be free). This should be straightforward.

Testing-wise, the error sadly appears only when we try to compile the generated code.

A good way to make sure it is working is then to add a new example.

n-osborne:
Closed by #228

PR #228, opened 2024-05-17
Add support for custom ghost type as model

n-osborne:
This PR proposes an answer to #209 by adding a translation for `type_declaration`s in ortac-core and use it in ortac-qcheck-stm.

The translation doesn't support constructors with inlined records, as they are buggy in Gospel anyway and it is still an open question whether they will be fixed or removed there.

The use of translated gospel types in ortac-qcheck-stm is naive: the symbols for constructors or record fields as translated are they appears in the gospel ast and are not stored in the context (as it is the case for gospel functions).


shym (approved):
LGTM, thanks!

Just a tiny comment:
- 90034f5 _Factorize logic of core_type_of_ty_  
  The modification to the .mli should be in the previous commit.
  Maybe the auxiliary common function could be called `core_type_of_ty_aux` or `core_type_of_ty_common` instead of `_core_type_of_ty`?

n-osborne:
CI failure is unrelated. Merging.

PR #229, opened 2024-05-17
Fix field access translation

n-osborne:
This PR fixes how the name of the accessed field is translated.

It is based on #228, please consider only the last four commits.


n-osborne:
Rebased on main as #228 has been merged

PR #230, opened 2024-05-21
Add support for functional models

### LINKED ISSUE #212: Parsing mistake of arrows from Gospel to Ortac in case of function declaration
When declaring the gospel model as a function `(string -> string option)`, Ortac misinterprets it and separates the operator `->`. It considers the arrow as a type constructor instead of the OCaml arrow, which is used for functions.

We have the type `t` modeling its content by a map of type `string -> string option`.
Following is the code generated by Ortac.
```
[...]
  type nonrec state = {
      contents: (string, string option) (->) }
[...]
```
It should be :
```
[...]
  type nonrec state = {
      contents: (string -> string option) }
[...]
```

n-osborne:
This PR proposes to add support for functional models in Ortac.

This fixes #212


nikolaushuber (approved):
LGTM

PR #234, opened 2024-06-05
Add empty cmd error

### LINKED ISSUE #213: Ortac could possibly generate a file without any commands
As a new user of Gospel and Ortac, I would have found it helpful to be able first to write an almost empty file with only `type definition` and `make function` without any commands. For now, Ortac requires at least one command to generate a valid OCaml file.
Considering the following interface file.
```
type t
(*@ model bla: integer *)

val make: unit -> t
(*@ t = make u
    ensures t.bla = 0 *)
```
Because Ortac does not detect the absence of commands, it generates an incorrect file with an unclear error.
```
9 |     type nonrec state = {
        ^^^^
Error: Syntax error
```
Ortac should crash before if it does not meet all the requirements in the interface instead of generating an invalid file.

Nevertheless, it might be nice to allow users to produce a very first Ortac-generated file with only `make` to ensure everything is well set up, even if Ortac is doing nothing. What do you think?

nikolaushuber:
This PR adds a new error which is triggered when there are no commands produced during the translation from Gospel to OCaml. 

This fixes #213

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # 0.3.0
 
+- Add error for empty command type 
+  [\#234](https://github.com/ocaml-gospel/ortac/pull/234)
 - Move to a module-based configuration
   [\#214](https://github.com/ocaml-gospel/ortac/pull/214)
 - Add support for custom ghost types as model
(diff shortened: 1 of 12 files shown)


n-osborne (approved):
Thanks for this nice looking PR! CI failure is unrelated.

Issue #182, opened 2023-11-20
Enable test of function without any sut as argument


n-osborne:
For now, these functions are ignored (and a warning is displayed to the user). Allowing to test these functions would increase test coverage.
This would also be a first step in making the tool more flexible.

n-osborne:
Fixed by #235

PR #235, opened 2024-06-10
Add support for functions without SUT argument

nikolaushuber:
This PR adds support for specifying functions that do not interact with the SUT. 

This responds to #182.


n-osborne (approved):
Nice work! Thanks!

I just have two tiny remarks, feel free to ignore:

- as already told offline, I prefer the tests to be run on every commit and not splitting the changes of the code and the conges on the tests.
- this one is really a matter of taste: I find `Option.fold` to be rather verbose compared to, and not clearer than, a simple pattern matching, so I tend to use the latter. This is particullary true when the term on each branch is rather complicated like:
  + [here](https://github.com/ocaml-gospel/ortac/pull/235/files#diff-cce013aa680978eb255a8a24cfccfa7f52526bc4d44fbd86627df8a94b39415dR152)
  + [here](https://github.com/ocaml-gospel/ortac/pull/235/files#diff-cce013aa680978eb255a8a24cfccfa7f52526bc4d44fbd86627df8a94b39415dR337)
  + [here](https://github.com/ocaml-gospel/ortac/pull/235/files#diff-cce013aa680978eb255a8a24cfccfa7f52526bc4d44fbd86627df8a94b39415dR416)
  + and [there](https://github.com/ocaml-gospel/ortac/pull/235/files#diff-cce013aa680978eb255a8a24cfccfa7f52526bc4d44fbd86627df8a94b39415dR467)

Issue #203, opened 2024-02-21
Add support for function returning tuples


n-osborne:
Functions returning tuples are filtered out in `Ir_of_gospel.no_functional_arg_or_returned_tuple`
Then it is possible to follow where the code need to be amended as there are some `failwith` talking about tuples.

jmid:
I hand-wrote support for a 6-tuple when I first needed it as part of testing weak hash sets:
https://github.com/ocaml-multicore/multicoretests/blob/7b9b1c123286f1f4db7d13e773f68198ca9b04cd/src/weak/stm_tests_hashset.ml#L112-L122

IIUC, it should be possible to generate similar lines (on demand) from the plugin, if we encounter a returned (or argument?) tuple. :thinking:

shym:
As `QCheck.Print.tup`_n_ are defined up to 9 only, I wondered whether it’s worth the effort of generating them rather than writing them once and for all.
(We might still be gentle and include in the generated file only the required tuples, though)

jmid:
True, tuples in OCaml APIs are relatively rare - and a 2-digit one more so! :smiley:
Down the line we may want to extend the support further, e.g., to record types, where a print function is not available.
At that point, ppx_deriving show may be an option.

jmid:
I believe this one is fixed by #237

n-osborne:
I believe you're right!

PR #237, opened 2024-06-19
Add support for tuples

nikolaushuber:
This PR adds support for tuples as arguments and as return values. 

This responds to #203


n-osborne (commented):
Thank you, very nice looking PR!
I just have a few remarks.

I really liked that you've splitted the PR in two steps, adding support for tuples as argument, then as returned values. This really eases the review :-)

I believe you didn't run the tests on every commit.

I suggest you fix the qualify bug (40edf86) before adding the test case that reveals it, as it generates incorrect code.

Now, for per commit remarks:

9b10630

You modified the documentation (the `example_limitations.mli` file) and the code. I think you didn't intend to as 1674f3c is dedicated to update the documentation.

It also contains some new test cases that would fit better in the commit adding the test in question (1d4dfda)

3efe5cfc3

You've added a new warning and filter out functions based on tuple arity in the `ir_of_gospel` phase. Which is a very good idea. This leads to removing a check you've just added in the previous commit (9b10630). Going in the reverse order, you wouldn't have to implement the removed check.

Also, the name is (rightfully) changed from `no_functional_arg_or_returned_tuple` to `no_functional_arg_or_big_tuple` later (5bda4ddc). I believe it would make more sense to change the function name in the same commit than its definition.

5bda4ddc

[`last_e`](https://github.com/ocaml-gospel/ortac/pull/237/files#diff-cce013aa680978eb255a8a24cfccfa7f52526bc4d44fbd86627df8a94b39415dR1012) is not a very informative name, may I suggest `ty_show` as it is what is being build?

[Here](https://github.com/ocaml-gospel/ortac/pull/237/files#diff-cce013aa680978eb255a8a24cfccfa7f52526bc4d44fbd86627df8a94b39415dR1021), you uses `QCheck.Print.tupX`. The rest of the generated code (and STM in general) uses `Util.Pp.pp_*` which exposes `pp_tupleX` for X from 2 to 10 (included). Maybe e should stick to the latter, for consistency sake.

Also, I think it would be better to use `qualify` rather than `evar` on a string containing dots (for consistency sake again, as `evar` uses `Longident.parse` internally which I've discovered recently and I'm now wondering if we should get rid of our `qualify`... but that would be a story for another PR anyway).

In [`tuple_types`](https://github.com/ocaml-gospel/ortac/pull/237/files#diff-cce013aa680978eb255a8a24cfccfa7f52526bc4d44fbd86627df8a94b39415dR1057), you can include the access to the returned type in your `aux`iliary function. This way you don't have to allocate the `ret_tys` list (not that ortac will be run on huge interfaces).

n-osborne (approved):
Thank you!

PR #239, opened 2024-06-20
Use existing functionaly from `Ppxlib.Longident`

n-osborne:
Self-explanatory.


nikolaushuber (approved):
Someone should eventually update/extend the documentation on Ppxlib ;)

Issue #238, opened 2024-06-19
[qcheck-stm] Incomplete computation warning 


nikolaushuber:
There seems to be a problem with printing of return values: 

```bash
$ cat > incomplete_comp.mli << EOF
> type 'a t
> (*@ model contents : 'a list *)
>
> val make : unit -> 'a t 
> (*@ t = make ()
>    ensures t.contents = [] *)
>
> val size : 'a t -> int 
> (*@ s = size t 
>   ensures s = List.length t.contents *)
> EOF
$ cat > incomplete_comp_config.ml << EOF 
> open Incomplete_comp 
> let init_sut = make () 
> type sut = int t
> EOF 
$ ortac qcheck-stm incomplete_comp.mli incomplete_comp_config.ml > /dev/null
File "incomplete_comp.mli", line 8, characters 0-81:
 8 | val size : 'a t -> int 
 9 | (*@ s = size t 
10 |     ensures s = List.length t.contents *)
Warning: Incomplete computation of the returned value in the specification of size. Failure message won't be able to display the expected returned value.
```

n-osborne:
Thanks!

I believe the problem comes from [here](https://github.com/ocaml-gospel/ortac/blob/main/plugins/qcheck-stm/src/ir_of_gospel.ml#L175)

Gospel sometimes add an `fs_apply` symbol in the AST that is not handle here.

PR #240, opened 2024-06-20
Remove bug with integers in return values 

nikolaushuber:
This PR removes a bug/shortcoming within the qcheck-stm plugin when testing functions that return integers. 
Gospel has two different types for integers (int & integer), but only int has a pretty-printer defined in STM. 
Gospel also automatically adds casting operations for ints to integers where needed. 

The code added in this PR dynamically detects where these casts have been added and automatically adds reverse-castings for error pretty-printers where needed. 

This responds to #238.


n-osborne (commented):
Thank you for this PR!

This looks really good.

I rather have the changes in the tests in the same commit than the changes in the code that cause them. I know that this generate a lot of "noise" in the diff, but we can always give `git` a path to focus on the part we want to inspect. And that way, we have coherent commits.

I've given some thoughts on how to handle the pretty printing of `Gospel.integer`s.
In your very detailed commit message (ddcdb3d7)(thanks for that :-)) you describe two ways:

> - Add integer pretty-printers to STM
> - Automatically cast integers back to ints when needed

The problem I see with the second option is that `integer_of_int` can raise an exception that is not caught in the generated code. Even, if we catch it, we are loosing some information when the `integer` is larger than an `int` (not sure it will happen that often though).

`Gospel.integers`s are implemented as `Z.t`, I think it would be worth it to ask the STM devs whether it could be intersting to add `Z.t` support. If not, there is also the solution to add them in the generated code (in the `Ty` module). We can add them by default: there is a rather high probability that the tested module contains some functions returning an `int` (casted by Gospel as an `integer`). In the case there is not, the cost is not prohibitive.

n-osborne (commented):
Very nice :-)

I just made a few located remarks.

n-osborne:
Thanks!

PR #245, opened 2024-07-22
Sut as type inst

nikolaushuber:
Fixes a bug were we are not checking for the sut type being used as the instantiation of a type variable, so for example 

```ocaml
val test : 'a t list -> int
(*@ r = test ts
    ensures r = List.length ts *)
```

would type check correctly and create incorrect code.


plugins/qcheck-stm/src/ir_of_gospel.ml
```diff
@@ -76,27 +76,35 @@ let ty_var_substitution config (vd : val_description) =
   let value_name, value_type = (vd.vd_name.id_str, vd.vd_type) in
   assert (is_a_function value_type);
   let open Ppxlib in
+  let open Reserr in
   let ret = function None -> Reserr.ok [] | Some x -> Reserr.ok x in
+  let check_sut_in_type_var ty =
+    match ty.ptyp_desc with
+    | Ptyp_constr (_, args) when List.exists (fun t -> Cfg.is_sut config t) args
```
n-osborne, line 83: This doesn't check deep enough. For example it won't detect `sut list list`.

nikolaushuber: True, I will adapt that!

plugins/qcheck-stm/src/ir_of_gospel.ml
```diff
@@ -91,11 +115,13 @@ let ty_var_substitution config (vd : val_description) =
               Reserr.(
                 error (Multiple_sut_arguments value_name, value_type.ptyp_loc))
         else aux seen r
-    | Ptyp_tuple elems ->
-        if List.for_all (fun t -> not (Cfg.is_sut config t)) elems then ret seen
-        else Reserr.(error (Returning_sut value_name, ty.ptyp_loc))
+    | Ptyp_tuple _ ->
+        let* _ = check `Right ty in
```
n-osborne, line 119: Same here

plugins/qcheck-stm/src/ir_of_gospel.ml
```diff
@@ -91,11 +115,13 @@ let ty_var_substitution config (vd : val_description) =
...
+        ret seen
     | Ptyp_constr (_, _) ->
-        if not (Cfg.is_sut config ty) then ret seen
+        if not (Cfg.is_sut config ty) then
+          let* _ = check `Right ty in
```
n-osborne, line 123: Same here.

plugins/qcheck-stm/src/ir_of_gospel.ml
```diff
@@ -76,11 +76,35 @@ let ty_var_substitution config (vd : val_description) =
...
+        | None ->
+            let* _ = List.map (check pos) elems |> sequence in
+            ok ())
+    | _ -> ok ()
+  in
   let rec aux seen ty =
     match ty.ptyp_desc with
     | Ptyp_any | Ptyp_var _ -> ret seen
     | Ptyp_arrow (_, l, r) ->
+        let* _ = check `Left l in
```
n-osborne, line 107: The expression evaluates to a `unit reserr`, so as well have `let* () =` rather than a wild pattern.

n-osborne (approved):
Great job, thanks!

Just had some tiny remarks, but after that it is good to merge.

PR #247, opened 2024-08-15
Add support for multiple sut arguments

nikolaushuber:
This PR lifts a previous restriction where only up to one sut argument can appear in any function to be tested. 
This works by keeping a stack of suts, which is created at startup with the required amount of sut arguments on it. Suts that are used as arguments are returned to the stack afterwards.


n-osborne (commented):
Thank you very much for this very nice looking PR!
We've already talked about the design in private conversations, so I mainly have some small comments on the programming style.

n-osborne (approved):
Thank you.

When reading the generated code, I was thinking it would be nice to have some documentation for the modules `SUT.Make` and `Model.Make` in `Ortac_runtime_qcheck_stm`. Even if we are not supposed to use these modules, it could help someone willing to understand the generated code (hence the trust one have in the generated tests).

n-osborne:
CI failure is unrelated to the PR.

Do you want to harmonise `ortac_cmd_show` and `ortac_show_cmd` names before merging?

n-osborne:
Thanks.
Merging.

PR #249, opened 2024-09-03
Fix invariant bug with multiple suts

nikolaushuber:
While allowing multiple suts in #247 a bug was introduced which causes the type invariants of the model not to be checked anymore after a field is modified. This PR fixes that.


n-osborne (approved):
Nice! Thank you!

`Stm_of_ir.translate_postcond.translate_invariants` begin to have a lot of parameters (two of type `Ident.t` amongst them). It could be useful document them for future readers.

n-osborne:
Also, what's the motivation behind moving to `Sequence.t` for the models in the invariant tests? I don't have anything against it, but some indication of the reason why in the commit message would be nice :-)

nikolaushuber:
> Nice! Thank you!
> 
> `Stm_of_ir.translate_postcond.translate_invariants` begin to have a lot of parameters (two of type `Ident.t` amongst them). It could be useful document them for future readers.

That is a good idea. Should I just add a comment or name the parameters? 

> Also, what's the motivation behind moving to `Sequence.t` for the models in the invariant tests? I don't have anything against it, but some indication of the reason why in the commit message would be nice :-)

The main motivation was that the Gospel standardlib apparently doesn't offer an append/concat function for lists, which I needed for the transfer function. Or maybe I missed it? But I will add some comment in the commit message :)

n-osborne:
Reading again the `test/invariants.mli`, the invariant is getting a bit weird with the presence of `transfer`.
Maybe someday we'll have a better idea of an example...

But for now, bug is fixed, CI is green. I'm merging. Thanks again :-)

PR #250, opened 2024-09-05
Some refactoring of `Reserr`

n-osborne:
This PR proposes some refactoring that I have accumulated in these last few months. Now that I have enough of them, let's try to merge them :-)

As it is purely refactoring, the tests, and more precisely the expect tests, are not modified. The tool is supposed to generate the same code. At some point I'd like to add some tests for this module, but in a future PR.


n-osborne:
Rebased. Ready for review.

n-osborne:
Thanks for taking the time to have a look!

Good catch for the unused `Warnings.kind`s, we've forgot to remove them when they became useless.

As this PR is already cleaning up the `Reserr` module, there is no need to cook another PR for that. I've removed them. I've also renamed `Returning_sut` to `Returning_nested_sut` for clarity.

I'll wait for the CI to finish and I'll merge.

PR #259, opened 2024-09-26
Add support for running tests in separate processes with a timeout

### LINKED ISSUE #254: [qcheck-stm] Running tests causing segmentation faults
When running a test with qcheck-stm that causes a segmentation fault there is not output to the user about which series of function calls lead to this scenario. 

Currently the only option is to manually adapt the call to `Test.make` in `/plugins/qcheck-stm/src/runtime/ortac_runtime_qcheck_stm.ml` to use `Util.fork_prop_with_timeout`. Maybe there could be a commandline flag to use this automatically, so that this manual adaptation is not needed?

nikolaushuber:
This fixes #254 

When a test causes a segmentation fault this results in the overall test suite process to crash. With this PR, the runtime looks for the presence of an environment variable, in which case it runs all tests in separate processes with a timeout. This allows to get traces both from crashing and non-terminating tests.

This also adapts the `dune` plugin to accept a new command-line flag with which the timeout can be set.

This is a bit of a WIP, since I could not find a good way of testing this feature. As far as I know the CI expects all tests to run without error, so I don't know how to make it accept an expected error.


nikolaushuber:
Rebased

n-osborne (approved):
Thank you! This looks very good!

I just have two small questions:

- `dd2f255` what happen when `ORTAC_QCHECK_STM_TIMEOUT` is set to a negative value?
- `c916ec7` maybe we don't need to be that specific on the cause of the crashes?

nikolaushuber:
> * `dd2f255` what happen when `ORTAC_QCHECK_STM_TIMEOUT` is set to a negative value?

That is indeed an interesting question. I just looked, under the hood it uses `Unix.alarm`, and the manpage for `alarm` does not state anything about what happens when you give it a negative value as argument ... If you try it, at least on my system it immediately returns with a non-zero return value. 

Should we adapt the code to in case of negative values also just use the default 10 seconds?

> * `c916ec7` maybe we don't need to be that specific on the cause of the crashes?

I can remove the first part about how segmentation faults can happen! I just thought it might be interesting for some people, especially since OCaml (if used without tricks 😅) promises to be a memory-safe language.

nikolaushuber:
> As a user, I think I would prefer it to fail rather than choosing a default value for me.

I made it so that anything that cannot be converted to a positive integer raises a failure now! 

> Maybe you can keep the Obj.magic related information as a side note, as example of crashing programs. I also like the idea of poping your third paragraph on top of the section, as it is a really good explanation of the problem the feature solves.

I adapted it accordingly 😃

plugins/qcheck-stm/doc/index.mld
```diff
@@ -470,3 +470,45 @@ and peculiarities:
...
+  qcheck-stm.sequential
+  qcheck-multicoretests-util
+  ortac-runtime
+  example)
+ (action
+  (setenv
+   ORTAC_QCHECK_STM_TIMEOUT
+   10
+    (run %{test} --verbose)))
+]}
```
jmid, line 512: One can also set environment variables on the command line and have them be visible in dune:
```
$ ORTAC_QCHECK_STM_TIMEOUT=10 dune test ...
```
I think that offers a reasonable workflow when one hits an unexpected segfault - just rerun the last command in the console with a small addition :slightly_smiling_face:

n-osborne: Indeed! Maybe we should distinguish between the CLI workflow and the CI workflow?

nikolaushuber: When I updated the documentation I was actually considering if I should mention other ways of setting environment variables, or if the standard user would already know about those. But I will add some additional comments about that!

n-osborne:
Thanks!

Regarding how to fail when the timeout is not a positive integer (`fcead56`), maybe we can do something cleaner than `failwith`.
I was thinking something along the way of `printf` followed by an `exit` with some error code (1 being the catch-all could be enough)

n-osborne:
Excellent. Thanks! Merging.

PR #261, opened 2024-10-03
Update Ortac/QCheck-STM Readme with found bugs

n-osborne:


jmid:
LGTM, great to have a growing list of example findings! :tada: 

Two observations:
- This is meant to document to new-comers that the tool is useful in practice. As such, we might consider adding it to the front-page README, where it is more likely to be seen? :thinking: 
- It is not an implementation bug, but the `Hashtbl.create` weirdness https://github.com/ocaml/ocaml/issues/13469 has been acknowledged in https://github.com/ocaml/ocaml/pull/13474#issuecomment-2388687681 agreeing that the documentation should be adjusted.

n-osborne:
I missed the evolution of the `Hashtbl.create` issue! Thanks!
I'll add it.

Regarding the question in which README to put the list, I was indeed wondering.

My train of thought was that the list is specific to Ortac/QCheck-STM.

But (1) maybe other plugins will find bugs in the future and a unique list may be preferable and (2) even `ocaml.org` is displaying the root README for `ortac-qcheck-stm` plugin (I was mistakenly thinking the opposite here).

PR #265, opened 2024-10-09
Add missing upper bounds on ortac packages

n-osborne:


n-osborne:
GH CI failing is unrelated (see: https://github.com/ocaml/setup-ocaml/issues/872)

PR #271, opened 2024-10-30
Extend tests of the QCheck-STM plugin

nikolaushuber:


n-osborne (approved):
Thanks! That's a nice addition. The more test, the merrier :-)

Regarding `Queue.take` (and similar examples)

```
+val take : 'a t -> 'a
+(*@ v = take t
+    raises Empty -> t.contents = old t.contents = Sequence.empty
+    modifies t.contents
+    ensures t.contents = match Sequence.length (old t.contents) with
+        | 0 -> Sequence.empty
+        | _ -> Sequence.tl (old t.contents)
+    ensures v = Sequence.hd (old t.contents)
+    ensures old t.contents <> Sequence.empty *)
```

The last postcondition will be violated in the same cases that the previous one will fail with a `Partial_function` exception. I would say the specifications are a bit verbose. But dealing with exceptional behaviour in Gospel for Ortac is not very nice in the first place.

Now I'm wondering:

- Do we want to force an evaluation order? For now, the two postconditions are arguments of `Ortac_runtime.append`, so there are no guarantees on evaluation order.
- should we consider boolean terms which evaluation raises `Partial_function` equivalent to `false`? This would make the last postcondition subsumed by the penultimate one.

nikolaushuber:
Indeed, I am actually not even sure why the last `ensures` clause is necessary. I think I copied that from another example that is already somewhere in the tests. But it does seem a bit redundant ... 

> Do we want to force an evaluation order? For now, the two postconditions are arguments of Ortac_runtime.append, so there are no guarantees on evaluation order.

IMO forcing an evaluation order for logical specifications seems a bit unnatural. In the end we would like to see the different clauses just as logical formulas, right?

> should we consider boolean terms which evaluation raises Partial_function equivalent to false? This would make the last postcondition subsumed by the penultimate one.

I'm actually unsure how that is currently handled? In case that the postcondition function raises `Partial_function` is that in itself not already a violation of the contract, as the evaluation of the specific formula is then undefined?

n-osborne:
> IMO forcing an evaluation order for logical specifications seems a bit unnatural. In the end we would like to see the different clauses just as logical formulas, right?

This is very true!

> I'm actually unsure how that is currently handled? In case that the postcondition function raises `Partial_function` is that in itself not already a violation of the contract, as the evaluation of the specific formula is then undefined?

For now, the `Partial_function` exception is printed using [`Printexc.register_print](https://github.com/ocaml-gospel/ortac/blob/main/src/runtime/ortac_runtime.ml#L125).

My feeling is that this is not very informative for the user.

But I'll open an issue on this particular problem, as the PR is good to be merged. Thanks again!

PR #272, opened 2024-11-02
Add command shrinker function

### LINKED ISSUE #269: [qcheck-STM] Emit a function to shrink `cmd` arguments
QCheck-STM will by default only shrink `cmd list`s. 
In practice, this means that, e.g., the random argument `-5` passed to `Hashtbl.create (-5)` will not be shrunk to the minimal `Hashtbl.create (-1)` before being printed.

To enable argument shrinking one has to write/emit a `shrink_cmd` function and pass it to `arb_cmd`.
Here's an example from multicoretests:
https://github.com/ocaml-multicore/multicoretests/blob/160394955526794ab5349ce683d9b7664d43bcb9/src/weak/stm_tests.ml#L33-L47
```ocaml
  let shrink_cmd c = match c with
    | Length -> Iter.empty
    | Set (i, d_opt) -> Iter.map (fun i -> Set (i,d_opt)) (Shrink.int i)
    | Get i      -> Iter.map (fun i -> Get i) (Shrink.int i)
    | Get_copy i -> Iter.map (fun i -> Get_copy i) (Shrink.int i)
    | Check i    -> Iter.map (fun i -> Check i) (Shrink.int i)
    | Fill (i,j,d_opt) ->
      Iter.(map (fun i -> Fill (i,j,d_opt)) (Shrink.int i)
            <+>
            map (fun j -> Fill (i,j,d_opt)) (Shrink.int j))

  let arb_cmd s =
    [...]
    QCheck.make ~print:show_cmd ~shrink:shrink_cmd [...]
```

- When no shrinking is needed (e.g., if there are no arguments or only a unit arg) one can just return the empty iterator `Iter.empty`
- When there's only one argument, one can use `Iter.map` over a built-in shrinker combinator from QCheck, e.g., of `int`, `char`, `string`, ...
- When there's more than one argument, we should shrink each argument separately. The `Fill` above case does this by composing two iterators with `<+>` (a shorthand for `Iter.append`). Why? Because a shrinker of `i` may return the empty iterator, if `i` is already minimal, e.g., `0` and we don't want the shrinker to only reduce when both `i` and `j` can be reduced. Their shrinking should be independent.

nikolaushuber:
This PR adds support for creating a function to automatically shrink the arguments of tested functions when creating counter-examples. 

This fixes #269.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Add command shrinker function
+  [\#272](https://github.com/ocaml-gospel/ortac/pull/272)
 - Remove unnecessary stack access from precondition
   [\#264](https://github.com/ocaml-gospel/ortac/pull/264)
 
--- a/examples/lwt_dllist_spec_tests.ml
+++ b/examples/lwt_dllist_spec_tests.ml
@@ -81,10 +81,24 @@ module Spec =
       | Take_r -> Format.asprintf "protect (fun () -> %s <sut>)" "take_r"
       | Take_opt_l -> Format.asprintf "%s <sut>" "take_opt_l"
       | Take_opt_r -> Format.asprintf "%s <sut>" "take_opt_r"
+    let shrink_cmd cmd__002_ =
+      let open QCheck in
+        let open Shrink in
+          let open Iter in
+            match cmd__002_ with
+            | Create () -> map (fun () -> Create ()) (unit ())
+            | Is_empty -> empty
+            | Length -> empty
+            | Add_l a_1 -> map (fun a_1 -> Add_l a_1) (int a_1)
+            | Add_r a_2 -> map (fun a_2 -> Add_r a_2) (int a_2)
+            | Take_l -> empty
+            | Take_r -> empty
+            | Take_opt_l -> empty
+            | Take_opt_r -> empty
     let cleanup _ = ()
     let arb_cmd _ =
       let open QCheck in
-        make ~print:show_cmd
+        make ~print:show_cmd ~shrink:shrink_cmd
           (let open Gen in
              oneof
                [(pure (fun () -> Create ())) <*> unit;
@@ -96,10 +110,10 @@ module Spec =
                pure Take_r;
                pure Take_opt_l;
                pure Take_opt_r])
-    let next_state cmd__002_ state__003_ =
-      match cmd__002_ with
+    let next_state cmd__003_ state__004_ =
+      match cmd__003_ with
       | Create () ->
-          let s__005_ =
+          let s__006_ =
             let open ModelElt in
               {
                 contents =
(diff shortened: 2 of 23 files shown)


jmid:
Wow Nikolaus, you are on fire! :fire: 

Have you observed whether this change significantly affects the time one has to wait to get a response (like the increased response time caused long arrays)?

Unrelated: Will the output shrinker reduce list arguments by suitable lifting of an element shrinker, as in `Shrink.(list int)`?

nikolaushuber:
> Have you observed whether this change significantly affects the time one has to wait to get a response (like the increased response time caused long arrays)?

Since all the tests run in milliseconds since ocaml-multicore/multicoretests#472 I didn't notice any change when running the tests, they still all run in under 1 second.

> Unrelated: Will the output shrinker reduce list arguments by suitable lifting of an element shrinker, as in Shrink.(list int)?

It did, but incorrectly so. I just ran it on-top of #271, which uses `of_list` in the array example. I pushed a small fix, since the argument shrinker to `Shrink.array` and `Shrink.list` is a labelled argument.

jmid:
Cool, nice catch :smiley: :+1:

n-osborne:
Thank you!
I'll try to take a look this week.

n-osborne (commented):
Thanks! Nicely done!

I just have a remark on how `shrink_cmd_cas` is defined in 8c50ae7.

You filter `list` and `array` to be able to apply the shrinker with a labelled argument, according to the `QCheck.Shrink` signature.

Note that `bytes` and `string` also ask for an optional shrinker argument.

Now, if users want to override one of these shrinker, they have to be careful to follow the `QCheck.Shrink.***` signature, which is reasonable, but should be documented.

Regarding shrinkers for types defined in the tested library, it should also be documented that users should go for unlabelled arguments.

One way to simplify the user's life would be to override on our side shrinkers with optional argument with shrinker with unlabelled arguments (*i.e*: `array shrink = array ~shrink`), as `shrinker_of_ty` is returning a shrinker, not an optional shrinker anyway. This should be done either by writing a module in the runtime that will be included in the generated code or directly in the generated code.

nikolaushuber:
> Note that bytes and string also ask for an optional shrinker argument.

That's true! I was considering to set the argument shrinker for `bytes` and `string` as well, but they already have a standard shrinker (i.e. the standard `Shrink.char`). I thought that it's probably very rare that a user would like to overwrite that with something else (e.g., `Shrink.char_printable` or so). 

> Regarding shrinkers for types defined in the tested library, it should also be documented that users should go for unlabelled arguments.

Indeed, that should be documented. In the current implementation it is also more or less required to define a shrinker for any custom defined types, as we do not inspect the config module. I guess one could make the custom shrinker function is the config module an optional value, but that probably gets ugly quite quickly as well .... 

Regarding the place for redefining the `Shrink` module, do you have a preference of either putting it in the runtime or in the generated code? IMO it's probably a bit cleaner in the runtime as it does not overcrowd the generated code.

n-osborne:
> Regarding the place for redefining the Shrink module, do you have a preference of either putting it in the runtime or in the generated code? IMO it's probably a bit cleaner in the runtime as it does not overcrowd the generated code.

I agree, the runtime is cleaner.

n-osborne:
@nikolaushuber will you be willing / have time to resume working on this PR ? (If no, I can update it and merge it for the next release).

Also, I'm not so sure about what I wrote:

> One way to simplify the user's life would be to override on our side shrinkers with optional argument with shrinker with unlabelled arguments (i.e: array shrink = array ~shrink), as shrinker_of_ty is returning a shrinker, not an optional shrinker anyway. This should be done either by writing a module in the runtime that will be included in the generated code or directly in the generated code.

Maybe it will bring more / another type of confusion.

Issue #270, opened 2024-10-24
[qcheck-STM] Add support for pure, first-class functions


jmid:
In the [Gospel paper](https://theses.hal.science/LRI-VALS/hal-02157484v2) *"GOSPEL -Providing OCaml with a Formal Specification Language"* I see in Sec.2.3 "Higher-Order Functions"
> "we cover in this paper only the simpler case where f is a pure function. In this case, we specify map as follows: [...]
> We leave the generalization to the general case of an effectful f to future work."

We should be able to support this with QCheck/STM too.

An example from multicoretests:
https://github.com/ocaml-multicore/multicoretests/blob/fa3f668e45e86995adb0e7dbb8ed0f26772ac503/src/lazy/stm_tests.ml#L36-L70
```ocaml
  type cmd =
    | [...]
    | Map of int_fun
    | Map_val of int_fun
  and int_fun = (int -> int) fun_

  let pp_int_fun par fmt f =
    Format.fprintf fmt (if par then "(%s)" else "%s") (Fn.print f)

  let pp_cmd par fmt x =
    let open Util.Pp in
    match x with
    | [...]
    | Map x -> cst1 pp_int_fun "Map" par fmt x
    | Map_val x -> cst1 pp_int_fun "Map_val" par fmt x

  let show_cmd = Util.Pp.to_show pp_cmd

  let arb_cmd _s =
    [...]
    QCheck.make ~print:show_cmd
      (Gen.oneof
         [ [...]
          Gen.map (fun f -> Map f)     (fun1 Observable.int int').gen;
          Gen.map (fun f -> Map_val f) (fun1 Observable.int int').gen;
         ])
```

Notes:
- The initial recursive type above is not needed. It is a left-over artifact from when we used `ppx_deriving.show` to derive print functions
- QCheck's function printers do not necessarily produce syntactically correct OCaml code
- For testing something like `val Hashtbl.fold : ('a -> 'b -> 'c -> 'c) -> ('a, 'b) t -> 'c -> 'c` more type instantiation (of `'c`) will be needed. Here one could just go with a default (int? char? ...) or instantiate to the same as `'a` or `'b` or ... :thinking:

jmid:
The lack of higher-order function support is the main blocker for achieving higher test coverage:

![API-coverage](https://github.com/user-attachments/assets/2a46bf3a-a70b-458c-bfc0-9f8546dc19f7)

jmid:
Fixed in #277

PR #275, opened 2024-12-03
Fix Unkown_kind

jmid:
Spotted while hacking on function generation support #270 

It triggered my ocd and hence require prompt fixing! :sweat_smile:


n-osborne:
Thanks!

PR #276, opened 2024-12-03
Remove left-over multiple-ts limitation from QCheck-STM README

jmid:
IIUC, this limitation was lifted by #247 and #253


n-osborne:
Yes, it was. Thanks for spotting that :-)

PR #277, opened 2024-12-05
Add function generation support 

jmid:
This (draft) PR add function generation, fixing #270

The core commit is 7a9df34 describing the change as:
    - building a generator with `QCheck.fun1-4` of `arbitrary.t` type
      that takes `Observable`s and an `arbitrary.t` result
      and finally selecting its `.gen` field to get a resulting `Gen.t`
    - quantifying `arbitrary.t` generators with `QCheck` to not pick up `Gen.t` ones
    - emitting the special `(_ -> _) fun_` type for an arrow type
    - invoking functions `f` with `QCheck.Fn.apply f` in both `run_cmd` and `next_state`
    - validating the fragment of functions we support, i.e., second-order and arity<=4

The PR is good enough shape to share and get feedback on at this point, since I'm starting to hit issues that are getting further from function generation itself.

One remaining issue is warnings about incomplete computations, which shows up in two places, despite both representing a complete model computation, that just happens to involve a function AFAICS:
```diff
+File "all_warnings.mli", line 39, characters 0-89:
+39 | val functional_argument : ('a -> bool) -> 'a t -> bool
+40 | (*@ b = functional_argument p t *)
+Warning: Incomplete computation of the returned value in the specification of functional_argument. Failure message won't be able to display the expected returned value.
```
and
```diff
-File "example_limitations.mli", line 23, characters 15-25:
+File "example_limitations.mli", line 23, characters 0-65:
 23 | val for_all : ('a -> bool) -> 'a t -> bool
-                    ^^^^^^^^^^
-Warning: Skipping for_all: functions are not supported yet as arguments.
+24 | (*@ b = for_all p t *)
+Warning: Incomplete computation of the returned value in the specification of for_all. Failure message won't be able to display the expected returned value.
```

Another issue which took some time to diagnose is a mixture of `List` and `Sequence` in the hashtbl specification.
In `filter_map_inplace` Ortac fails to parse the specification correctly, which results in `next_state` modeling a state change, despite the `modifies h` specification.
The issue is that `h.contents = Sequence.filter_map` is parsed as
```
TApp ( "infix =", [ TApp ( "Sequence.of_list", [...] ),  ... ] )
```
i.e., with nested applications on the left-hand-side of the equality sign `=`.

Commit 2d0ce5a adds a fix attempt for it, which makes Ortac output a `next_state` function honoring the `modify h` clause, however the produced code then fails to type-check because of the mixture of `List` and `Sequence`:
```ocaml
$ git checkout 2d0ce5a
[...]

Issue #224, opened 2024-04-18
[wrapper] Add support for models


n-osborne:
Gospel is moving towards talking only of logical models in the specifications (not OCaml values).

Ortac/QCheck-STM is already using relational postconditions to propose model-based testing (maintaining an implementation of the logical model in parallel of the OCaml value).

There are some clearly identified limitations on taking this approach for Ortac/Wrapper (see section 7.1 of [Clément Pascutto PhD dissertation](https://www.pascutto.fr/media/dissertation.pdf)

I propose to take another approach with Ortac/Wrapper, inspired by bisimulation (see also Section 5 of [this paper](https://dl.acm.org/doi/10.1145/581690.581696))

A bisimulation is a property on labeled transition systems.

We can see function calls (and the corresponding relational postconditions) as such labels.

We then want to check that:

```
   a   ~  m
   |      |
 φ |      | σ
   V      V
   a' ~   m'
```

Where:
- `a` and `a'` are OCaml values (`a'` correspond to `a` after the function call)
- `m` and `m'` are logical models (`m'` is `old m` in Gospel terms)
- `φ` is an OCaml function call
- `σ` is the modification of logical models expressed in a relational postcondition
- `~` is the relation between OCaml values and logical models that defines the bisimulation and can be implemented with a projection (a `to_model` function): `a ~ m ≡ to_model a = m`

Obviously, we can't generate the projection for the user (we mainly deal with abstract types).

Following the same idea as #214, I propose that the user provides these projections in a module-based configuration, relying on a naming convention.

Then, when/before translating the Gospel terms, we will substitute access to a logical model (it is coded as the access to a field in current Gospel) with a call to the right projection.

Whenever a projection is not available, we can skip the Gospel term.

That way, given an interface like this one (in `queue.mli` file):

```ocaml
type 'a t
(*@ mutable model contents : 'a list *)

val create : unit -> 'a t
(*@ q = create ()
    ensures q.contents = [] *)

val push : 'a -> 'a t -> unit
(*@ push a q
    modifies q.contents
    ensures q.contents = a :: old q.contents *)
```

Ortac/Wrapper would generate something along the line of:

```ocaml
include Queue

let contents = .... (* taken from the configuration module *)

let create () =
  let q = create () in
  assert (contents q = []);
  q

let push a q =
  let q' = copy q in
  push a q;
  assert (contents q = a :: contents q')
```

PR #279, opened 2024-12-10
Replace lists by sequences in Ortac/QCheck-STM tests

### LINKED ISSUE #278: Rewrite all tests using `list`s as model with `sequence`s
The future new gospel standard/logical library won't contain the `List` module.

n-osborne:
Fixes #278

This should also help #277.

I didn't change the tests for the wrapper mode as using sequences would mean supporting models (see #224)

### DIFF
--- a/plugins/qcheck-stm/test/all_warnings.mli
+++ b/plugins/qcheck-stm/test/all_warnings.mli
@@ -1,13 +1,13 @@
 (*@ predicate p (x : 'a) = true *)
 
 type 'a t = { v : 'a array }
-(*@ mutable model contents : 'a list *)
+(*@ mutable model contents : 'a sequence *)
 
 type s
 
 val make : int -> 'a -> 'a t
 (*@ t = make i a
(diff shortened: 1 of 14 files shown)


jmid:
LGTM! 

Will `::` be a short-hand for `Sequence.cons` in the revised stdlib, that removes `List`? 
Comparing `tuples.add` it helps readability, IMO:
```diff
- ensures h.contents = match tup with a, b -> (a, b) :: old h.contents
+ ensures h.contents = match tup with a, b -> Sequence.cons (a, b) (old h.contents)
```

n-osborne:
Sadly, I'm not sure.
It is not yet decided whether we will continue to allow importing pure OCaml value in Gospel. If so, `::` will be reserved for OCaml lists. Is not, an argument can be make that allowing `::` for Gospel sequences can be confusing?
But I agree that the prefix `cons` is not ideal.

TBH, I'm also a bit sad that we lose pattern matching too.

PR #283, opened 2025-01-06
Add x-maintenance-intent field in opam files

n-osborne:


hannesm:
Thanks for this PR, unfortunately the syntax is different that what we proposed. You use `("latest")`, we specify `"(latest)"`. Would you mind to change this (and move the quotes outwards one character)? Thanks a lot.

n-osborne:
Of course!
Thanks for pointing it out.
Do I need to update the opam files on ocaml/opam-repository too?

hannesm:
> Do I need to update the opam files on ocaml/opam-repository too?

yes, that'd be lovely

Issue #294, opened 2025-03-20
[wrapper] Violated pre and postconditions should raise `Violated_condition`


Lucccyo:
In the specification of the function `euclidean_division`, the precondition `requires y > 0` enforces that the second argument must be strictly positive:

```ocaml
val euclidean_division: int -> int -> int * int
(*@ q, r = euclidean_division x y
    requires y > 0
    ensures  x = q * y + r
    ensures  0 <= r < y *)
```
When this precondition is violated, the wrapper mode of Ortac generates an error to inform the user. These errors are defined in `Ortac_runtime`.

However, when a precondition (or postcondition) is violated, the program should raise `Violated_condition` instead of `Violated_invariant`, as shown in the following generated code:

```ocaml
  if
    not
      (try
         Ortac_runtime.Gospelstdlib.(>)
           (Ortac_runtime.Gospelstdlib.integer_of_int y)
           (Ortac_runtime.Gospelstdlib.integer_of_int 0)
       with
       | e ->
           ((Ortac_runtime.Specification_failure
               { term = "y > 0"; term_kind = Pre; exn = e })
              |> (Ortac_runtime.Errors.register __error__001_);
            true))
  then
    (Ortac_runtime.Violated_invariant { term = "y > 0"; position = Pre }) |>
      (Ortac_runtime.Errors.register __error__001_);
```

PR #295, opened 2025-03-20
Fix incorrect error type for violated preconditions and postconditions

Lucccyo:
When a precondition or postcondition is violated, Ortac generates a raise to `Violated_invariant` from the `Ortac_runtime`, which is incorrect (as explained in [#294]). The expected behaviour is to raise `Violated_condition` instead, as preconditions and postconditions are contract conditions, not invariants.

This PR fixes this by matching `kind` with the expected error type when `kind` is violated.
```ocaml
let violated kind = match kind with
  | `Pre | `Post -> violated_condition (term_kind kind)
  | _ -> violated_invariant (term_kind kind)
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+- Fix incorrect error type for violated preconditions and postconditions
+  [\#295](https://github.com/ocaml-gospel/ortac/pull/295)
+
 # 0.6.0
 
 - Make search for returned value description more flexible
--- a/plugins/monolith/test/generated/wrapper.expected.ml
+++ b/plugins/monolith/test/generated/wrapper.expected.ml
@@ -108,8 +108,8 @@ module R =
                   |> (Ortac_runtime.Errors.register __error__012_);
                 true))
       then
-        (Ortac_runtime.Violated_invariant
-           { term = "0 <= n <= 32"; position = Pre })
+        (Ortac_runtime.Violated_condition
+           { term = "0 <= n <= 32"; term_kind = Pre })
           |> (Ortac_runtime.Errors.register __error__012_);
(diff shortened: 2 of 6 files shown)


n-osborne:
Thanks!
I haven't yet take a look at the modification but you need:
1. to promote the changes in the expect tests
2. add an entry in the Changelog (in a separated commit)

n-osborne:
Thanks again for this PR!

Overall, we now have the expected behaviour.

I only have two suggestions

I think the `Failure.violated` that you redefined to make it handle correctly the `ensures` and `requires` clauses is still a bit fragile. Its internal logic dictates that everything that is not a pre or a post condition is an invariant. The whole code works as expected because we never call the function on something else than `ensures` and `requires` clauses. In particular, we use directly `Failure.violated_invariant`. I suggest to do the same for `Failure.violated_condition`, which will allow to remove `Failure.violated`.

On the topic of cleaning up, we never use the `Failure.\`Invariant` polymorphic variant. I believe it is safe to remove it totally and it makes sense as an invariant is checked as a pre and post condition anyway.

Lucccyo:
Thanks. I agree that we should handle the violated errors separately to remove `Failure.violated` completely. However, this function is also used for an `xpost` case in the function `with_xposts`, and I am unsure what error should be raised.
```ocaml
        (fun (p, t) ->
          let s = term_printer t in
          let nonexec exn = F.spec_failure `XPost ~term:s ~exn ~register_name in
          term ~context nonexec t
          |> Result.map (fun t ->
                 case ~guard:None
                   ~lhs:(xpost_pattern ~context name p)
                   ~rhs:
                     [%expr
                       if not [%e t] then
                         [%e F.violated `XPost ~term:s ~register_name]]))
```

Issue #201, opened 2024-02-14
[wrapper] Add support for `old` operator


n-osborne:
This issue proposes to add support for the `old` operator in the `wrapper` plugin.

For now, terms containing an `old` operator are simply skipped.

Clément has already some code [here](https://github.com/pascutto/ortac/tree/old) using the marshalling technic described in his thesis and the paper [here](https://hal.science/hal-03690675v2). That could be a very good starting point.

There is some rebasing to do to begin with.

n-osborne:
Closed by #297

PR #297, opened 2025-03-31
Add support for `old` operator in the wrapper plugin

Lucccyo:
Resuming Clément Pascutto's work to complete the implementation of the old operator in the wrapper plugin.
This PR aims to address #201.


n-osborne (commented):
Thanks! Good work adapting the patch.

There is a bug in `Ir_of_gospel.old_down` in the `Tlambda` case. It calls `old_vars` on the body of the lambda where it should make a recursive call. There is no test-case for it so it doesn't show, but you can trigger it with a dummy postcondition without any `old` operator and involving an anonymous function with one of the header's variable in its body.

I do believe the PR should be split into at least four separated commits (plus the changelog update). This would help understand what's going on and document some choices in the commit messages.

The four topics I see are:

- code cleanup (remove the mutable field in the IR)
- collect the copies (most of the work)
- implement the runtime copy function
- uses the copies in the generated code (also update the tests)

IIUC, `Ir_of_gospel.collect_old` is doing two things:
1. replace terms under an `old` operator with a variable
2. collect all these new variables
It is not obvious from the function's name and I don't see any better naming choice at the moment. So maybe a comments could be useful?

It could also be worth it to document a bit `Ir_of_gospel.old_vars` and  `Ir_of_gospel.old_down`.

As a nitpick, there is no consistency with the `Tterm.` prefix, sometimes it is present, sometimes not. Lots of them are unnecessary and could be removed.

n-osborne (approved):
Thanks, this looks very good.

PR #299, opened 2025-04-14
Add tests to wrapper plugin

Lucccyo:
Testing cases where the generated instrumented OCaml program should fail.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Extend Wrapper plugin tests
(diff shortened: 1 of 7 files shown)


n-osborne (approved):
Thanks for this PR!

I believe you can squash the two first commits (promoting the tests in the same commit that causes the changes).

We also usually put a capital letter in the commit messages in this project.

The `suite` directory was used for this kind of tests (testing the behaviour of the generated code). But I think it is OK to do it this way too.

PR #301, opened 2025-05-05
`lwt_dllist_spec.mli` should meet the implementation of `lwt_dllist`

Lucccyo:


n-osborne:
Thanks! This looks nice.

You should add a lower bound on `lwt-dllist` in the description of `ortac-examples` in the `dune-project` file as this won't compile for previous versions.

Regarding the commit message, I wouldn't say `mutable` is a prefix but rather keyword. Also, a conjunction in a commit message is usually a hint that this could have been two commits :-)

Lucccyo:
Thanks for the review !

PR #305, opened 2025-05-12
Add support for models in the wrapper plugin

Lucccyo:
WIP:
This PR adds support for models in the Wrapper plugin, using ideas from #224.
To enable model support in Wrapper, a projection function should be implemented and exposed in the interface file. This function should be named the same as the gospel model it targets and annotated in the interface file with the attribute `[@@model]`.
When multiple models are defined, the same number of projection functions should be provided.

Internally, these projection functions are handled as regular `values` except that we register if it has the attribute `model` or not.

For example, the projection function `view`:
 ```ocaml
type 'a t
(*@ mutable model view : 'a sequence *)

val view : 'a t -> 'a list [@@model]
```
During translation, `collect_model` replaces all occurrences of the gospel model with calls to the corresponding projection function (both in the postconditions and invariants).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- [Wrapper] Add support for models in the wrapper plugin
+  [\#305](https://github.com/ocaml-gospel/ortac/pull/305)
 - [QCheck-STM] Disable warning 34 in generated code
   [\#307](https://github.com/ocaml-gospel/ortac/pull/307)
 - [Wrapper] Reorganise tests into separate files
--- a/plugins/wrapper/src/ir.ml
+++ b/plugins/wrapper/src/ir.ml
@@ -62,6 +62,7 @@ type value = {
   arguments : ocaml_var list;
   returns : ocaml_var list;
   register_name : string;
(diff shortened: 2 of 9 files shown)


n-osborne (commented):
Thanks, this is a very nice looking PR!

I just have three remarks.

- First commit:

(nitpick) In `Ir_of_gospel.is_projection`, it is good style to name the auxiliary function.

In `Ir_of_gospel.type_.process`, line 650, you don't need the `let` binding as the function immediately returns the bounded expression.

- Second commit:

As discussed offline, in `Ir_of_gospel.collect_model`, line 37, the way the function application is build is a bit strange (updating the `ls_constr` field to true whereas it is not a constructor). In fact, in the current Gospel, function applications are represented with the help of the `fs_apply` symbol. `f a b` is represented as `Tapp (fs_apply, [ f; a; b])`. This is bound to change with the next release of Gospel (as it is a arguably not the best design), but it may be worth it to try to be correct *vis-à-vis* the current version to avoid bad surprises.

n-osborne (approved):
Thanks! Merging.

PR #323, opened 2025-06-11
Rename wrapper tests to make them more readable

Lucccyo:


n-osborne:
`Behaviour` is indeed a strange name. Why not simply `reference` as the content is about a reference.
IIUC, you are testing two separate things here:
- Ortac/Wrapper generate compiling code
- Ortac/Wrapper generate code actually checking the specification at runtime (what you refer to as behaviour I believe)
I would then suggest to name the library from there content, and rename `test` as `behaviour_test` or `runtime_assertion_checking_test` or something similar.

n-osborne:
Conflicts should be resolved in order to merge this PR.

Issue #320, opened 2025-06-11
[Wrapper] Fix naming conflict between model names and argument names


Lucccyo:
Suppose two arguments in different functions exist in a Gospel environment with the same name (for example, several functions have `t` as an argument). In that case, a number will suffix everyone following the first to make it unique (the second will be `t_1`).
If the model's name is the same as a Gospel argument (previously declared because, with some luck, the name could be valid), it leads to an error.

Consider the following example where the model `n` stands for the line number of the current Pascal triangle.
In the generated wrapper file, the model name of `n` is correct.
The translation of `comb_aux` has `n` as an argument, and `comb` has `n_1` as an argument.
When checking the conditions of the functions, we have `r.n` in the Gospel contract, which should be translated to `n r` because we know that `n` is a model of `r`. However, the following error is raised.

```ocaml
112 |           (Ortac_runtime.Gospelstdlib.integer_of_int (n_2 r)) =
                                                            ^^^
Error: Unbound value n_2
```

In my opinion, the `r.n` in Gospel makes the `n` kind of an argument, so it is translated with the suffix of the number because the previous one was `n_1` for the `comb` function.

```ocaml
type pascal
(*@ mutable model n: int
    mutable model row: int list *)

(*@ function rec comb_aux (c: integer) (y: integer) (i: integer) (k: integer) : integer =
  if i > k then c else comb_aux (c*y/i) (y-1) (i+1) k *)

(*@ function comb (y: integer) (k: integer) : integer = comb_aux 1 y 1 k *)

val n : pascal -> int [@@model]
val row : pascal -> int list [@@model]

val init : unit -> pascal
(*@ r = init ()
    ensures r.n = 0 *)

val next : pascal -> unit
(*@ next r
    modifies r
    ensures r.n = old r.n + 1
    ensures forall i. 0 <= i <= r.n -> r.row[i] = comb r.n i
    *)
```

We should avoid this error because it can lead to the generated source file not working without explanation.
To correct it, I suggest naming the models with a unique name similar to invariants (`__invariant___001_`) and storing a map between model names and their unique names.

Lucccyo:
See #324.

PR #324, opened 2025-06-13
Fix naming conflicts between model projections and function arguments

Lucccyo:
This PR addresses issue #320 by separating projection functions and regular values in the intermediate representation.
Each model projection function is now assigned a unique name (e.g., `__projection_modelname_001`) to prevent naming conflicts with existing Gospel identifiers such as function arguments. 
In order to ensures that references to models in invariants are translated in calling the projection functions, these projections functions are emitted before everything including invariants in the generated file.


n-osborne (commented):
Thanks for this PR!

First commit doesn't compile: it introduces a pattern on `Projection` which doesn't exist yet. I believe it can be safely push later in the history.
You should promote the tests at every commit, not in a separated one. Adding some test-cases can be done in a separated commit (it is even a good idea).

The triplet in the `models` field in `Ir.type_` could benefit from some documentation, either with a comment or using a record instead.

PR #332, opened 2025-06-23
Add a better management of attribute for projection functions

Lucccyo:
This PR introduces new naming rules for the projection functions to reduce the need for explicit `[@@model]` attributes:
- The attribute is not required if the function has the same name as a model.
- Otherwise, the user must explicitly specify the model using `[@@model "model_name"]`.

For example, the following type `t` has two models `capacity` and `contents`. This involves two projection functions, `capacity` wich has no attributes as it has the same name as the gospel model. Nevertheless, `to_list` is the projection function of `contents` and needs the attribute `[@@model "contents"]`.

```ocaml
type t
(*@ mutable model capacity: int
    mutable model contents: int list *)
val capacity : t -> int
val to_list : t -> int list [@@model "contents"]
```

This simplifies the syntax in common cases while preserving clarity in ambiguous situations. The user is also free to put a custom name for the projection function.


n-osborne:
Do we really need the attribute's payload to be a string ?
Wouldn't is be cleaner to ask for:

```ocaml
type 'a t
(*@ model contents : 'a sequence *)

val to_list : 'a t -> 'a list [@@model contents]
```

Or even:

```ocaml
val to_list : 'a t -> 'a list [@@model_for contents]
```

Or even further (this one introduce a new word though):

```ocaml
val to_list : 'a t -> 'a list [@@projection_for contents]
```

Lucccyo:
I agree that the payload doesn't need to be a string as long as it type-checks!
Also, I like your suggestion to make the attribute name more expressive, especially the last option:
`[@@projection_for contents]`. It conveys the intention behind the annotation.

Lucccyo:
I suggest accepting both the "string style" and the "OCaml identifier style" to avoid some errors or to be too strict with the user. What do you think @n-osborne ?

PR #336, opened 2025-06-25
Prepare release 0.7

Lucccyo:


plugins/wrapper/README.md
```diff
@@ -78,15 +78,67 @@ very beginning:
...
+val capacity : 'a t -> int
+val to_list : 'a t -> 'a list [@@projection_for contents]
+```
 
-- `model`s,
-- the `old` operator.
+Here, there is no ambiguity that `capacity` corresponds to the model `capacity`,
+but `to_list` needs to be explicitly declared as the projection for the model `contents`.
 
-[main README]: ../../README.md#supported-gospel
```
n-osborne, line 92: We need to keep the link to point to the description of the translatable subset of Gospel terms.

plugins/wrapper/README.md
```diff
@@ -78,15 +78,67 @@ very beginning:
 module Lib = LibAsserts
 ```
 
+## Advanced specification
 
-## Supported Gospel
+The wrapper plugin also supports advanced Gospel features, such as
+type models and `old` statements,
```
n-osborne, line 84: logical models?

n-osborne (commented):
Just some comments on the READMEs

plugins/wrapper/README.md
Lucccyo: Ok !

plugins/wrapper/README.md
```diff
@@ -78,15 +78,69 @@ very beginning:
...
 
-## Supported Gospel
+The wrapper plugin also supports advanced Gospel features, such as
+logical models and `old` statements,
+allowing more precise and expressive specifications for abstract types.
+Let's consider a toy example of a polymorphic container type with limited capacity:
 
-The wrapper plugin has currently some limitations on what Gospel
-specifications are supported. Apart from the general restriction to
-the executable fragment of Gospel (as mentioned in the [main README]),
```
n-osborne, line 86: We also need to keep the information about the general restrictions. You have lifted the imitations coming for the wrapper part of Ortac/Wrapper, but we still have the limitations coming from ortac-core.

n-osborne:
The main README needs to be updated to, as it mentions Ortac/Wrapper as an experimental (as in not-yet-released) plugin.
This is also a good opportunity to put a word about Ortac/Dune which is missing from the README.

n-osborne:
CI failure is unrelated. Merging.

PR #348, opened 2025-07-21
Fix cycle dependency in `Ortac/Dune-rules` for wrapper plugin

Lucccyo:
During the library declaration in the `Ortac/dune-rules`, the libraries stanza should include `ortac-runtime` and the original library (e.g., `my_lib`) instead of the wrapped one.
```
  (library
   (name named_wrapper)
   (modules named_wrapper)
-   (libraries ortac-runtime named_wrapper)
+   (libraries ortac-runtime my_lib)
   (package my_package))
```


Lucccyo:
The CI error appears to be not related to the work.

art-w (approved):
Looks good! Thanks for strengthening the cram test by calling dune to check that ortac generates valid rules :)

n-osborne:
Nice, thanks!
 Two minor remarks:
- bumping the version number in *monolith.opam files should be done in a separated PR
- commits should be reordered (fix->test->changelog, or test->fix-> changelog)

n-osborne:
Merging.
Thanks again!

